PR #5, opened 2016-06-09
Update class-pins.topojson

jgovea25:
Added my username to the San Francisco pin!

### DIFF
--- a/class-pins.topojson
+++ b/class-pins.topojson
@@ -19,7 +19,7 @@
       { "type": "Feature",
         "geometry": {"type": "Point", "coordinates": [-122.391269, 37.782140]},
         "properties": {"prop0": "value0"}
-        "properties": {"user2": "jgovea25"}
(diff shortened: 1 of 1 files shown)


brianamarie:
Hi @jgovea25 👋 

I was just checking out your pull request and was going to suggest an edit, but it looks like you've got it figured out! Nice! 👍  

Once this check passes, go ahead and merge! 🎉 🚢

brianamarie:
Hi @jgovea25 👋 

You probably notice this is showing a merge conflict. This is happening because the initial changes that were made caused Git to view the later changes as conflicting, even though the text is what it should be for our activity. 😄 

There are two ways you could fix this: 
1. If you're comfortable using Git locally, feel free to resolve the conflict on your own.
2. Close this pull request and start the activity again on a new branch. 

Please let me or any of the other trainers in the chat know if you want or need any help. 😄 ✨

Issue #27, opened 2016-06-10
Adding Portland, OR to the map


patriciaa:
## Adding Portland to the map

Hi everyone! This issue is to track the addition of Portland to the map, which can be found at:

- Longitude: 122.6765° W
- Latitude: 45.5231° N

brianamarie:
There are changes coming to this project and repository. To help get it ready, we are hoping to clean up some outstanding issues and pull requests. If you need any additional help, please feel free to open a new pull request. 😄 Thank you and happy coding! :octocat:

PR #28, opened 2016-06-10
created mymap.topojson w coordinates

patriciaa:
* added Portland, OR to the map
* fixes issue #27 
* @githubteacher


NickTikhonov:
Looks good, but I think there is a slight problem with the instructions - the map won't render with your pin until you remove `json` from the front of the file :)

patriciaa:
Thanks! Changed it :)

brianamarie:
> Looks good, but I think there is a slight problem with the instructions - the map won't render with your pin until you remove json from the front of the file :)

@NickTikhonov Great point - thank you! 😄

Issue #69, opened 2016-07-11
Adding Hood River, OR to the map


ktfeucht:
## Adding Hood River, OR to the map

Hi everyone! This issue is to track the addition of Hood River, OR to the map, which can be found at:

- Longitude: -121.521462 
- Latitude: 45.705397

crichID:
:wave: @ktfeucht it looks like your Pull request has been merged, but it did not close the issue. It is probably because your forgot to add the special key words GitHub was looking for or added extra words between the magic words and the issue number. The magic words include `fix`, `fixes`, `fixed`, `close`, `closes`, `closed`, `resolve`, `resolves`, `resolved` + the Issue number. So your PR would have looked something like this `This Pull Request resolves #69`. I will go ahead and close this issue since it looks like you have successfully added your pin to the map! 

Welcome to GitHub!

PR #85, opened 2016-07-11
Adding Hoor River, OR

ktfeucht:
## Hood River, OR Add

Adding Hood River to the map to close issue #69 

@githubteacher - Why?
* It's awesome.
* There is skiing November - August
* The mountain biking is rad
* So are the kite and windsurfing


lukeart:
Looks awesome, @ktfeucht !

PR #97, opened 2016-07-26
Sekti92 jkt

sekai92id:
just add few files on my branch. please take a look


sekai92id-jkt.topojson
```diff
@@ -0,0 +1,12 @@
+{ 
+  "type": "Feature", 
+  "geometry": 
+  {
+    "type": "Point", 
+    "coordinates": [106.845599, -6.208763]
```
sekai92id: now you know my location

sekai92id: now you know my location

Issue #209, opened 2016-08-09
My City Arlington Heights


Devansu:
## Adding <your-city> to the map

Hi everyone! This issue is to track the addition of <your-city> to the map, which can be found at:

- Longitude: -87.980627
- Latitude: 42.088360

PR #211, opened 2016-08-09
Dpatel arlington heights

Devansu:


Devansu:
Add my city Arlington Heights to class project to review by @githubteacher and close #209

PR #394, opened 2016-08-30
Add Tdotcdot88 location

tdotcdot88:
Adding my location to the map


brianamarie:
@tdotcdot88 Looks great! Make sure the file is named `_pins/Tdotc88.json` so it will appear in the right file, then feel free to 🚢!

PR #438, opened 2016-08-31
Consolidate files from extra teams

brianamarie:
👋 Hello! If you are being messaged about this pull request, it is likely because you created a 'team' and a new repository and were doing ✨  learning there instead of on this repository. Instead of asking everyone to re-do their work, I want to add your pins now. (That's why you get this notification! 😄 ) 

**You don't need to do anything - this is just to let you know that this is happening. **

You might notice that your other repository and team are going to go away, but we don't want your pin to be gone. You'll still see it here, along with all of the other pins that are displayed on the course home page. (Note: These changes won't take effect until I merge this pull request.) 

Those who _may_ be affected: 
- @hectorsector 
- @crichID 
- @hollenberry 
- @beardofedu 
- @AndriiGitHub
- @awoerner92
- @thinhdd13
- @cito125
- @raunaksingwi
- @lijessje
- @Mohit7th

Again, no action is required - this is more of a 'heads up' from GitHub trainers that we are making corrective action so your pins will be displayed properly.

PR #441, opened 2016-08-31
Move pins into _pins directory

hectorsector:
This PR moves all the pins into the proper directory so they're automagically rendered.


hectorsector:
Taken care of in #438

PR #469, opened 2016-09-01
add comment

osundblad:
Just some text


comment.txt
```diff
@@ -0,0 +1 @@
+This is my home town...
```
osundblad: @githubteacher

Issue #14, opened 2016-06-09
San Diego, CA


hollenberry:
## Adding San Diego to the map

Hi everyone! This issue is to track the addition of San Diego to the map, which can be found at:

- Longitude: -117.1294°
- Latitude: 32.7456°

The01Geek:
Adding Point Lomca, SD

luketrotta:
add Carlsbad, San Diego, CA USA

kerrystanko:
Adding my pin of North Park to the map!

githubstudent:
test

developmentim:
Adding another pin for North Park

PR #503, opened 2016-09-05
Add Carlsbad to map

luketrotta:


_pins/luketrotta.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: luketrotta
+latitude: 33.1581
+longitude: 117.3506
```
hollenberry: Almost there! I think that longitude needs to be - 😄

luketrotta:
closes #14

hollenberry:
This looks good to 🚢 !

PR #555, opened 2016-09-07
_pins/calhone.json

Calhone:
Add Telluride pin
@githubteacher 
close#a1a2299


Calhone:
Is the file name required twice?

PR #568, opened 2016-09-08
Update tycecyle.json

tycecycle:
Forgot lines 1 and 5 in original file.

### DIFF
--- a/_pins/tycecyle.json
+++ b/_pins/tycecyle.json
@@ -1,3 +1,5 @@
++---
 githubHandle: tycecycle
 latitude: 44.052069
 longitude: -123.086754
++---


_pins/tycecyle.json
```diff
@@ -1,3 +1,5 @@
++---
```
hollenberry: Hey @tycecycle ! 

I think that this might be error-ing because of a very minor syntax error. You were almost there!

Instead of `+---`, lines 1 and 5 should only include `---`. 

This may have happened on copy paste, as the files changed view will show you line additions signified by the + sign, but we don't want to actually add the + sign into our markdown document while editing.

tycecycle: Great thanks @hollenberry !

PR #593, opened 2016-09-08
Adexot lagos

pigeonn:


hollenberry:
Hey @pigeonn ! It looks like you might be using a format that we used previously, but have since updated from. 

Prior to merging, I'd recommend updating your syntax to match the most recent instructions 👍 

The most relevant can be found here: https://github.github.com/on-demand/intro-to-github/05/

Issue #487, opened 2016-09-03
Lukavac


perlba:
## Adding <Soda> to the map

Hi everyone! This issue is to track the addition of <Lukavac> to the map, which can be found at:

 Longitude:        44.533951
    Latitude:       18.532639

perlba:
:+1:

perlba:
Closed

perlba:
Adding again :+1:

PR #594, opened 2016-09-08
Adding to the map Lukavac #487

perlba:
Hi everyone! This the addition of SODA City to the map, which can be found at:


perlba:
My issue is  Lukavac #487 not not 594 can I change it ?

PR #601, opened 2016-09-09
Adding to the map

perlba:
Hi everyone!


ericmjalbert:
Hey there, it seems like your Lat and Long integer values should not be surrounded with "<" and ">". That's probably why Travis-CI is failing your checks.

Issue #473, opened 2016-09-01
issue goal


lunar9:
## Adding <your-city> to the map

Hi everyone! This issue is to track the addition of Philadelphia to the map, which can be found at:

- Longitude: <your-longitude>39.952584
- Latitude: <your-latitude>-75.165222

PR #607, opened 2016-09-09
Delete pinslunar9

mccurcio:
This is not where you live.


hollenberry:
I agree. This doesn't match up with #473 , where Philadelphia is the pin. Go ahead and 🚢 when ready.

mccurcio:
cleaning

PR #639, opened 2016-09-09
renamed to _pins folder

mccurcio:


mccurcio:
DH location was not in proper folder

PR #640, opened 2016-09-09
Delete id_rsa.enc

mccurcio:
just cleaning up


rturgano:
You can **DO IT!!!!** 
Image of encouragement

Issue #782, opened 2016-09-19
Adding Irvine, CA to the map


lynnlangit:
## Adding Irvine, CA to the map

Hi everyone! This issue is to track the addition of Irvine, CA to the map, which can be found at:
- Longitude: 33.6839° N
- Latitude: 117.7947° W

ericaz98:
Hi, I'll be adding my name to the same city.

blinebau:
Adding my name to Irvine as well.

PR #784, opened 2016-09-19
adding me again

lynnlangit:
fun tutorial!

closes #782

### DIFF
--- a/_pins/LynnLangit.json
+++ b/_pins/LynnLangit.json
@@ -1,0 +1,5 @@
+---
+githubHandle: LynnLangit
+latitude: 33.6839° N
+longitude: 17.7947° W
+---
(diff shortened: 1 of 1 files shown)


_pins/LynnLangit.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: LynnLangit
+latitude: 33.6839° N
+longitude: 17.7947° W
```
regnerjr: I'm thinking this formatting might cause some issues. 
The `.json` file that we create is a special type of javascript file, 
`json` is short for JavaScript Object Notation, and should probably be in only plain characters, 
I think just having the numbers i.e. `latitude: 33.6839` should fix the build issues. 
Good Luck. 👍

PR #812, opened 2016-09-21
my new maps add to maps

ctdavid:


hectorsector:
@ctdavid I'm closing this PR as I think you've got another one going in the right direction in #821

PR #949, opened 2016-09-29
add SteveC to pins for washington dc

steve-12XU:
**pull request test**

i created an intro file for the @githubteacher, as requested.

closes #949


steve-12XU:
Fixes #949

Issue #967, opened 2016-10-02
First Issue


winstonhoseadriggins:
## Adding <Chester, PA> to the map

Hi everyone! This issue is to track the addition of <Chester, PA> to the map, which can be found at:

- Longitude: <-75.355746>
- Latitude: <39.849557>

PR #968, opened 2016-10-02
Added coordinates of location

winstonhoseadriggins:


winstonhoseadriggins:
Added coordinates per @githubteacher request close #967

Issue #1005, opened 2016-10-04
Davis, CA


latomkovic:
## Adding Davis, CA to the map

Hi everyone! This issue is to track the addition of Davis, CA to the map, which can be found at:
- Longitude: 38.544907
- Latitude: -121.745366

tnelsen:
Add tsnelsen pin to map

PR #1006, opened 2016-10-04
Add latom pin to user map

latomkovic:


latomkovic:
I added **my pin** to the map!

@githubteacher this closes #1005

PR #1109, opened 2016-10-08
Create Saturn2357-Portland2

Saturn2357:


_pins/Saturn2357-Portland2
```diff
@@ -1,6 +1,6 @@
 ---
```
Aric22Bear: these lines should be removed, just keep three lines which contain your username, latitude and longitude, and then that will work

Saturn2357: That worked! :-) However my icon isn't showing up on the map after I have merged. Is there something I am doing wrong?

jr-cologne: The lines mustn't be removed. Put three of them to the top and three of them to the bottom.

Then your pin should show up correctly.

jr-cologne: Like that:

```
---
githubHandle: Saturn2357
latitude: 45.466388
longitude:-122.626374
---
```

Saturn2357: Once I put the lines back in place, Travis CI (which I have no clue about) rejects my merge once again. There must be some trick in getting the pin on the map. Did I miss anything on the tutorial?

PR #1141, opened 2016-10-10
Adding a pin to the map

Hkhan15:


Hkhan15:
@githubteacher **Activity: Adding a New Pin to the Map.** closes #1141

PR #1473, opened 2016-11-01
Add Dearborn

mossimo88:
Hello! 

I am adding Dearborn, MI pinned to the map. 

Thanks!

### DIFF
--- a/Dearborn
+++ b/Dearborn
@@ -1,0 +1,5 @@
+---
+githubHandle: mossimo88
+latitude:42.322260 
+longitude:-83.176314
+---


brianamarie:
Hi @mossimo88! Make sure that you have spaces in between the colon and the numbers, like this:

`latitude: numbers` instead of `latitude:numbers`

Dearborn
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: mossimo88
+latitude:42.322260
```
crichID: @mossimo88 please add a space after the `:` in the latitude and longitude.

Dearborn
```diff
@@ -0,0 +1,5 @@
+---
```
crichID: @mossimo88 you will need to add this file to the `_pins` directory and add the `.json` file extension for it to work!

PR #1474, opened 2016-11-01
Create okutao.json

ogacho:
Pinning Nairobi on the map
closes #2 Nairobi


ogacho:
syntax error corrected

PR #1553, opened 2016-11-07
Add PIN for Roanoke

bpary:
Created a new PIN for Roanoke, VA based on instructions from @githubteacher.


vyablonskyy:
Nice addition to the class. Good work. 👍

PR #1654, opened 2016-11-08
ぱみゅぱみゅ - つけまつける

rj60001:


_pins/reiss-jones-tokyo.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: rj60001
```
crichID: @rj60001 you will need to add the `---` lines at the beginning and end of your file so it looks like this:

```
---
githubHandle: rj60001
latitude: 139.770210
longitude: 35.725987
---
```

PR #1736, opened 2016-11-10
Create audrick.json

audrick:
add long lat in file


liemdoan:
You need to add the '---' characters to beginning and end of file. Otherwise looks good.

_pins/audrick.json
```diff
@@ -0,0 +1,4 @@
+githubHandle: audrick
```
crichID: @audrick please add the `---` at the beginning and end of the file! After that your tests should be passing and you will be able to :ship:.

PR #1794, opened 2016-11-16
Add latitude and longitude of Rösrath

pe-mo:
Added an introduction file for GitHub


crichID:
@pe-mo thank you for creating a pin. The tests were failing due to some incorrectly formatted pins that were merged before you created yours. We have fixed those pins and then merged master into your branch. Your test is passing now and you should be able to click the big green button below to **Merge pull request**.

PR #1798, opened 2016-11-16
Validate coordinates and YAML format

kylemacey:
This validates the coordinates in YAML against a regular expression. Additionally, if the YAML file is invalid, the test will raise an exception and fail.

Also fixed any failing "JSON" files

cc: @brianamarie @crichid @hectorsector @stoe


kylemacey:
This should be good for peer review 👍

hectorsector:
[This test case](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/commit/6677b8e46230dac1c1ad6ac5a19a848d357338cc) passes when it shouldn't. Should this be caught by the YAML linter, maybe?

crichID:
@kylemacey thank you for creating this new test! We just found another example this test _would_ have caught so I am going to go ahead and :ship: this! :heart:

kylemacey:
> This test case passes when it shouldn't. Should this be caught by the YAML linter, maybe?

@hectorsector the trailing dashes are optional in the YAML spec. The opening dashes are required.

Issue #1904, opened 2016-11-24
Des Moines, Iowa


murken:
## Adding Des Moines to the map

Hi everyone! This issue is to track the addition of Des Moines, Iowa, USA to the map. The city (state capital building) can be found at:

- Longitude: <-93.603880>
- Latitude: <41.591118>

PR #1905, opened 2016-11-24
Create murken.json

murken:
adding Des Moines pin to map


murken:
thanks @githubteacher.
closes #1904

PR #1973, opened 2016-11-29
adding stockholm to the map

darwonsamal:
Adding stockholm into the map because sweden.


X8invisible:
Make Stockholm great again

PR #1979, opened 2016-11-29
adding Belfast

Thebiglad:
I have added Belfast pin to the map because it is an amazing city :+1:


helenalucas96:
https://www.youtube.com/watch?v=9i1w4m_hnDw
FIRE

PR #1980, opened 2016-11-29
Create cleary-gfm.json

cleary-gfm:
added my lat and long :space_invader:


brianamarie:
Hi @cleary-gfm! I think the reason the test is still failing is because of 'trailing spaces' on line 3.

brianamarie:
🎉 🚢

PR #2031, opened 2016-12-01
Create gangov.json

gangov:
Those are the latitude and longitude of Plovdiv


dyhun:
Hi @gangov you need to add "---" these caracters to the first and last line, if you want check an exiting pin

PR #2066, opened 2016-12-05
Create eprattic.json

eprattic:
Add my latitude and longitude for my user @githubteacher


dyhun:
Hi @eprattic you nedd to add "---" these caracters at the first and last line ,check an existing pin

PR #2165, opened 2016-12-10
Stanleymwangi tokyo

stanleymwangi:


jr-cologne:
Good job. Now make sure to delete the branch and then everything is fine. :)

Issue #2212, opened 2016-12-13
Buffalo, New York


buffaloser:
## Adding <your-city> to the map

Hi everyone! This issue is to track the addition of Buffalo to the map, which can be found at:

- Longitude: 42.886447
- Latitude: -78.878369

PR #2215, opened 2016-12-13
Buffalo

buffaloser:
here's the location of Buffalo. @githubteacher closes #2213


buffaloser:
here's the location of Buffalo. @githubteacher closes #2212

PR #2243, opened 2016-12-14
Add username, url, date

monroeld:


crichID:
@monroeld I am so excited to see you are completing our next class on using GitHub locally. We actually use a different repository for that class since the activity is slightly different: https://github.com/githubschool/on-demand-github-pages. Please comment on [Issue 1](https://github.com/githubschool/on-demand-github-pages/issues/1) in that repository to be added as a collaborator and re-create this file there!

Issue #446, opened 2016-09-01
Denver


ArcticLlama:
## Adding Denver to the map

Hi everyone! This issue is to track the addition of Denver to the map, which can be found at:
- Longitude: 104.9903 W
- Latitude: 39.7392 N

tnordloh:
Howdy!  I'll be adding my name to the pin for Denver

sallarak:
Adding my name to the pin for Denver

AdrienneChristy:
Hi @tnordloh and @sallarak, I'll be adding my name too.

iamstevevan:
Mile High!

elvman12:
I am in Denver as well... woohoooo.

sbs205c5:
Also in Denver!

gllisch:
Denver!

PR #2255, opened 2016-12-15
Coordinates

gllisch:
Added this because its where I am here #446


gllisch.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: gllisch
```
gllisch: Is this the right name?

gllisch:
👍 Perfect

gllisch:
@githubteacher closes #446

gllisch.json
jr-cologne: Yes, that should be right.

gllisch:
Thanks @jr-cologne for the help

jr-cologne:
You are welcome. 👍

PR #2269, opened 2016-12-17
Create the file with fedorz coordinates

fedorz:


_pins/fedorz.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: fedorz
+latitude: 45.495473
+longitude: -122.578735
+---
```
fedorz: I don't know exactly what I'm doing, tbh.

PR #2295, opened 2016-12-17
added new city

### LINKED ISSUE #2285: Adding my favourite city-Taipei
## Adding <your-city> to the map

Hi everyone! This issue is to track the addition of <Taipei> to the map, which can be found at:

- Longitude: <25.0330>
- Latitude: <121.5654>

jojojolin:
Added taipei

closes #2285

### DIFF
--- a/jolin
+++ b/jolin
@@ -1,0 +1,3 @@
+githubHandle:jojojolin
+latitude:121.5654
+longitude:25.0330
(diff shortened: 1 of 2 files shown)


jolin
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle:jojojolin
+latitude:121.5654
+longitude:25.0330
```
jr-cologne: and here

leiguoyang:
@jojojolin I think you've confused the `latitude` value with the `longitude` value. In this case, your pin on the map may not display.

PR #2298, opened 2016-12-17
Create ysalo.json

ysalo:


_pins/ysalo.json
```diff
@@ -0,0 +1,5 @@
+---
+gitHubHandle: ysalo
```
jr-cologne: It's ``githubHandle``

_pins/ysalo.json
```diff
@@ -0,0 +1,5 @@
...
+latitude: 47.322322
+Longitude: -122.310682
```
jr-cologne: the ``L`` of ``Longitude`` has to be small

_pins/ysalo.json
africak: :+1:

_pins/ysalo.json
```diff
@@ -0,0 +1,5 @@
+---
+gitHubHandle: ysalo
+latitude: 47.322322
```
africak: lower case `l` for `latitude` :boat:

PR #2304, opened 2016-12-18
Add my pin

Ji3m:
Created the file Ji3m.json which contains my latitude and longitude for my residence in Dallas, TX so that a pin can be added to the map. Whoopi-do.

@githubteacher  My comment.

closes #2304


Ji3m:
Fixes #2304

PR #2348, opened 2016-12-21
New addition

smwq:
Add a new addition to bangalore.
new pull request.

### DIFF
--- a/_pins/smwq.json
+++ b/_pins/smwq.json
@@ -1,3 +1,5 @@
+---
 githubhandle: smwq
 latitude: 12.971599
 longitude: 77.594563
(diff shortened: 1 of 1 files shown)


_pins/smwq.json
```diff
@@ -0,0 +1,3 @@
+githubhandle: smwq
```
crichID: @smwq your file is almost finished. You just need to add a series of three dashes `---` as the top line and the bottom line of the file so it looks like this:

```
---
githubHandle: smwq
latitude: 12.971599
longitude: 77.594563
---
```

Please let me know if you need help!

PR #2371, opened 2016-12-22
Create ezeetester.json

ezeetester:
Add latitude and longitude. githubteacher. closes #3

### DIFF
--- a/_pins/ezeetester.json
+++ b/_pins/ezeetester.json
@@ -1,0 +1,3 @@
+githubHandle: ezeetester
+latitude:12.971599
+longitude:77.594563
--- a/_pins/ezeetester.json
+++ b/_pins/ezeetester.json
@@ -1,3 +1,4 @@
 githubHandle: ezeetester
 latitude:12.971599
 longitude:77.594563
+Favoritecolor: Green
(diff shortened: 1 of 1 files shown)


_pins/ezeetester.json
```diff
@@ -0,0 +1,4 @@
+githubHandle: ezeetester
```
crichID: @ezeetester you must begin and end the file with a line that contains only 3 dashes `---`.

_pins/ezeetester.json
```diff
@@ -0,0 +1,4 @@
...
+latitude:12.971599
```
crichID: @ezeetester you must include a space after the `:` for the latitude and longitude.

_pins/ezeetester.json
```diff
@@ -0,0 +1,4 @@
...
+longitude:77.594563
+Favoritecolor: Green
```
crichID: @ezeetester we are excited to know your favorite color is green - especially since that is the color of new lines of code added to the repository. Unfortunately, leaving this line will cause the test to fail, so you should probably remove it!

jnorris18:
Nice work!

PR #2428, opened 2016-12-27
create new json file

ahkeno:
@githubteacher


ahkeno:
can't create one githubHandle in two different location ?

PR #2466, opened 2016-12-29
TheBearMethod Coordinates

TheBearMethod:
Second attempt, @githubteacher  closes #2465

### DIFF
--- a/_pins/TheBearMethod.json
+++ b/_pins/TheBearMethod.json
@@ -1,0 +1,5 @@
+ ---
+ githubHandle: TheBearMethod
+ latitude: 56.0833
+ longitude: -3.4667
+ ---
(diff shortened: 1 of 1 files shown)


_pins/TheBearMethod.json
```diff
@@ -0,0 +1,5 @@
+ ---
```
jr-cologne: try to remove the space before the three lines

_pins/TheBearMethod.json
```diff
@@ -0,0 +1,5 @@
...
+ githubHandle: TheBearMethod
+ latitude: 56.0833
+ longitude: -3.4667
+ ---
```
jr-cologne: here too

_pins/TheBearMethod.json
TheBearMethod: I will try that, copy and pasted there version....

PR #2541, opened 2017-01-04
create json file with my current coordinates

arcaf:


arcaf:
# Adding my pin to the project map
This change will include my coordinates to be pinned in the project map  and closes #2541 

PR #2572, opened 2017-01-05
Add raffylorenzana's location coordinates

raffylorenzana:
Hi @githubteacher, I have added a file that contains my location coordinates

PR #2575, opened 2017-01-05
Add raffylorenzana's pin to the map

raffylorenzana:
@githubteacher Add my pin to the map
closes #2572


raffylorenzana:
@githubteacher Add my pin to the map

This pull request closes #2575

PR #2597, opened 2017-01-05
specified pin location

biggoof:
@githubteacher is this what you wanted?
This may have closed #2597, then again it may not have.


biggoof:
This may have closed #2597, then again it may nit have.

PR #2649, opened 2017-01-08
Create mtanduong.json

mtanduong:
Coordinates for mtanduong


kcmin:
@mtanduong try removing the spaces at the beginning of each line in your .json file

PR #2664, opened 2017-01-09
Create username

VTG38:
@VTG38

### DIFF
--- a/username
+++ b/username
@@ -1,0 +1,1 @@
+vtg38-Barre
--- a/username
+++ b/username
@@ -1,1 +1,6 @@
-vtg38-Barre
+VTG38  
+I'm just a mom, but a cool one learning to code.  
+Love to dance
+Humor is my best quality
+I work in healthcare
+


username
```diff
@@ -0,0 +1,6 @@
+VTG38
```
VTG38: [](url)From b2d1fc09ff14e764bb05c151d927bd52c7cd47a3 Mon Sep 17 00:00:00 2001
From: VTG38 <VTG38@users.noreply.github.com>
Date: Sat, 7 Jan 2017 21:11:44 -0500
Subject: [PATCH] Delete _config.yml

---
 _config.yml | 3 ---
 1 file changed, 3 deletions(-)
 delete mode 100644 _config.yml

diff --git a/_config.yml b/_config.yml
deleted file mode 100644
index ceaed8c..0000000
--- a/_config.yml
+++ /dev/null
@@ -1,3 +0,0 @@
-collections:
-  - pins
-exclude: [vendor]

PR #2765, opened 2017-01-11
added gargvk81 coordinates

gargvk81:
@githubteacher


gargvk81:
removed the extra dashes at the end.

PR #2859, opened 2017-01-16
Create tharbert.json

tharbert:
Added location.  @githubteacher


bobpurdon:
@tharbert looks good.

PR #2860, opened 2017-01-16
Create location identifier

tenwalls:
cos i can


georgepatterson:
@tenwalls Looks good 👍

PR #2866, opened 2017-01-16
Adding anastassia-s's coordinates

anastassia-s:
This pull request closes #2819.


highoctane305:
Hey there, see you're having some trouble.  If you add --- at the start of the file and on a line by itself at the end of the file you should be sorted,

anastassia-s:
@highoctane305 Thank you!!!

PR #2941, opened 2017-01-18
Added Nottingham with latitude and longitude

Rhyspacker:


samadmoizc8:
@Rhyspacker That's look great 👍

PR #2973, opened 2017-01-19
adding GitHubTeacher's coordinants

vaggkos:


_pins/githubteacher.vaggkos
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: vaggkos 
+Your Latitude: 37.9833
```
crichID: @vaggkos you will want to remove the `Your ` from before Latitude and Longitude so your tests will pass and you can merge your pull request!

PR #2978, opened 2017-01-19
Adding humanmirjalai's coordinants

humanmirjalali:
@githubteacher What do you think? Is everything ok?

### DIFF
--- a/humanmirjalali.json
+++ b/humanmirjalali.json
@@ -1,3 +1,5 @@
- githubHandle: humanmirjalali
- latitude: 47.6130
- longitude: -122.2008
+--- 
+githubHandle: humanmirjalali
+latitude: 47.613016
+longitude: -122.200816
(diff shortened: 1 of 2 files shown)


humanmirjalali.json
```diff
@@ -0,0 +1,5 @@
+ ---
+ githubHandle: humanmirjalali
+ latitude: 47.613092
```
humanmirjalali: This is latitude of my house.

humanmirjalali.json
```diff
@@ -0,0 +1,5 @@
...
+ longitude: 122.200885
```
humanmirjalali: This is longitude of my house.

rvbarreto:
The check is failing because there are spaces in the beginning of each line. When you remove then, it will pass the test!

PR #3014, opened 2017-01-21
add my handle's pin to the map

jkhashi:
I feel welcome!

### DIFF
--- a/_pins/jkhashi.json
+++ b/_pins/jkhashi.json
@@ -1,0 +1,5 @@
+---
+joe k.:jkhashi
+lattitude: 33.770050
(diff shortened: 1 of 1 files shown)


vt49701:
Great job!  Way to go!

flonach:
Good game !

_pins/jkhashi.json
```diff
@@ -0,0 +1,5 @@
+---
+githubhandle: jkhashi
+lattitude: 33.770050
```
crichID: @jkhashi one more small thing to fix. This should be `latitude` (one t after the a)!

PR #3035, opened 2017-01-22
Adding my pin to the GitHub User map

Runtime-Learner:
I *have* added my pin to the map (just like the tutorial asked).

Tring to use markdown formatting is
>Pretty fun
>pretty useful
> **REALLY** useful


Runtime-Learner:
Fixes #3035

PR #3112, opened 2017-01-24
Fixing syntax errors in pins

crichID:
@brianamarie this might be interesting to you!


brianamarie:
@crichID After looking into this just a little, I'm more confused than ever. 

I've viewed the repository at the point in time when the commit that passed the test was created, and all of the tests seemed right. I'm very confused at this!

PR #3221, opened 2017-01-28
@Githubteacher closes#3

srinikanan:
srini did some modification


_pins/githubRichardSpano.json
```diff
@@ -0,0 +1,5 @@
+```
```
crichID: @srinikanan these ````` should be three dashes instead `---`. If you fix this on the first and last line of the file, your tests will pass and you will be able to merge!

RichardSpano:
Changed the ticks to dashes in my .json

PR #3228, opened 2017-01-28
Add alandr0id pin

alandr0id:
Add alandr0id pin

### DIFF
--- a/_pins/alandr0id.json
+++ b/_pins/alandr0id.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: alandr0id
 latitude: -17.989356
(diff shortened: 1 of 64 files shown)


_pins/alandr0id.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: alandr0id
+latitude: -17.989356
+longitud: -67.137805
```
crichID: @alandr0id it looks like the test is failing because the `e` got cut off of the end of `longitude`. If you fix that everything should be passing and you will be OK to merge!

PR #3402, opened 2017-02-02
add wwx's pin into the map

wenxingwang:
adding WWX into map

### DIFF
--- a/_pins/wwx.json
+++ b/_pins/wwx.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: wwx
-latitude:51.472488
-longitude:-0.974936
+latitude: 51.472488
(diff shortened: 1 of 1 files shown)


_pins/wwx.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: wwx
+latitude:51.472488
```
crichID: @wenxingwang you just need to add a space after each : and your tests should pass!

crichID:
@wenxingwang I just made a quick comment on what you need to fix so your checks will pass. Here is a video that covers how to make these changes if it is helpful: https://vimeo.com/201011288

wenxingwang:
merge completed

PR #3340, opened 2017-02-01
(WIP) Limit map pins

stoe:
## :construction: WIP :construction:

**:exclamation: Do NOT merge :exclamation:**

As described in https://github.com/github/services-web/issues/655#issuecomment-276651827, this PR adds limiting the number of pins returned in `createMap.topojson`.

Merging now won't have an effect right now as our pin files don't provide a `date` property needed for the sorting to work.

PR #3429, opened 2017-02-03
Add YAML front matter validation

swinton:
Per [my comment](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/pull/3340#issuecomment-277053698), we stand to benefit from some YAML front matter validation, as we have some data issues currently.

This PR adds a test, to ensure all the `_pins/*.json` files begin and end with [the appropriate front matter formatting](https://jekyllrb.com/docs/frontmatter/).

We should also clean up [the existing problematic files](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/pull/3340#issuecomment-277053698) before merging.

### DIFF
--- a/tests/test_coordinateValidation.rb
+++ b/tests/test_coordinateValidation.rb
@@ -16,11 +16,4 @@ class TestCoordinateValidation < Test::Unit::TestCase
 
   end
 
-  def test_frontmattervalidation
-      Dir["./_pins/*.json"].each do |path|
-          lines = File.readlines(path)
-          assert_match(/^-+$/, lines[0])
-          assert_match(/^-+$/, lines[-1])
-      end
-  end
 end
--- a/tests/test_coordinateValidation.rb
+++ b/tests/test_coordinateValidation.rb
@@ -15,4 +15,12 @@ class TestCoordinateValidation < Test::Unit::TestCase
     end
 
   end
+
+  def test_frontmattervalidation
+      Dir["./_pins/*.json"].each do |path|
+          lines = File.readlines(path)
+          assert_match(/^-+$/, lines[0])
+          assert_match(/^-+$/, lines[-1])
+      end
+  end
 end
--- a/tests/test_frontmatterValidation.rb
+++ b/tests/test_frontmatterValidation.rb
@@ -8,13 +8,13 @@ class TestFrontMatterValidation < Test::Unit::TestCase
             data = YAML.load_file(path)
 
             keys = data.keys.map { |k| k.downcase }
-            assert_true(keys.include? "githubhandle")
-            assert_true(keys.include? "latitude")
-            assert_true(keys.include? "longitude")
+            assert(keys.include? "githubhandle")
+            assert(keys.include? "latitude")
+            assert(keys.include? "longitude")
 
             lines = File.readlines(path)
             assert_equal("---\n", lines[0])
-            assert_true(lines[4..-1].include? "---\n")
+            assert(lines[4..-1].include? "---\n")
         end
     end
 
--- a/tests/test_frontmatterValidation.rb
+++ b/tests/test_frontmatterValidation.rb
@@ -1,0 +1,21 @@
+require 'test/unit'
+require 'yaml'
+
+class TestFrontMatterValidation < Test::Unit::TestCase
+
+    def test_frontmattervalidation
+        Dir['./_pins/*.json'].each do |path|
+            data = YAML.load_file(path)
+
+            keys = data.keys.map { |k| k.downcase }
+            assert_true(keys.include? "githubhandle")
+            assert_true(keys.include? "latitude")
+            assert_true(keys.include? "longitude")
+
+            lines = File.readlines(path)
+            assert_equal("---\n", lines[0])
+            assert_true(lines[4..-1].include? "---\n")
+        end
+    end
+
+end
--- a/tests/test_frontmatterValidation.rb
+++ b/tests/test_frontmatterValidation.rb
@@ -8,13 +8,13 @@ class TestFrontMatterValidation < Test::Unit::TestCase
             data = YAML.load_file(path)
 
             keys = data.keys.map { |k| k.downcase }
-            assert(keys.include? "githubhandle")
-            assert(keys.include? "latitude")
-            assert(keys.include? "longitude")
+            assert(keys.include? "githubhandle", "githubhandle missing: #{path}")
+            assert(keys.include? "latitude", "latitude missing: #{path}")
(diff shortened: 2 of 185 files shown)


tests/test_coordinateValidation.rb
```diff
@@ -15,4 +15,12 @@ def test_coordinatevalidation
...
 
   end
+
+  def test_frontmattervalidation
+      Dir["./_pins/*.json"].each do |path|
+          lines = File.readlines(path)
+          assert_match(/^-+$/, lines[0])
+          assert_match(/^-+$/, lines[-1])
+      end
+  end
```
stoe: The test seems a bit restrictive.

What do you thing about this one, @swinton 

```rb
def test_frontmattervalidation
  Dir["./_pins/*.json"].each do |path|
    assert_match(/^-{3,3}$/, lines[0])
    assert_equal(
      File.readlines(path).grep(/^-{3,3}\n/).length,
      2
    )
  end
end
```

It checks if the `---` occurs in the first line and twice in the file.
As we have a YAML notation, there could be text after the second `---` the ☝️  tests respects that.

swinton: Good point, @stoe.

> As we have a YAML notation, there could be text after the second --- the ☝️ tests respects that.

I agree that forcing the last to end with `---` is overly restrictive.

However, it does seem we have a requirement for these `_pins/*.json` files to conform to a well-defined data structure:

```yaml
---
githubHandle: [a-zA-Z0-9] 
latitude: [float]
longitude: [float]
---
[text:optional]
```

And I think we need an additional 1 or 2 tests that ensure submissions conform to this data structure.

I'll try and cook something up.

stoe: > And I think we need an additional 1 or 2 tests that ensure submissions conform to this data structure.

:100: x agree.

I actually would prefer if we had a real JSON file, rather than a YAML with a `.json` extension, but unfortunately collections need to be in a text format with frontmatter to be consumable by Jekyll.

swinton: > I actually would prefer if we had a real JSON file, rather than a YAML with a .json extension

Yes. Should we rename the files to be `.md`? Under a separate effort, perhaps?

swinton:
I cleaned up some of the data, and added a test for front matter validation.

@stoe, could you take another look?

brianamarie:
@swinton The test here looks good - thank you for always jumping right in to fix problems that you see. ❤️ 
 
It was my understanding that the .yamllint tests were checking this content, but obviously that wasn't the case. Do you think we could/should adapt the [yamllint configurations](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/blob/master/.yamllint) ([documentation here](https://yamllint.readthedocs.io/en/latest/)) as an alternative?

swinton:
> Do you think we could/should adapt the yamllint configurations (documentation here) as an alternative?

@brianamarie, I don't see anything in [those docs](https://yamllint.readthedocs.io/en/latest/) that would allow us to specify which keys we are expecting to see in YAML submissions ([e.g.](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/blob/ba5fd6e1142a00dcc7b8abfddded7adde5b994d2/tests/test_frontmatterValidation.rb#L11)). Unless I'm overlooking something, I think we need [the new test](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/blob/ba5fd6e1142a00dcc7b8abfddded7adde5b994d2/tests/test_frontmatterValidation.rb#L6) to validate this, as I see some contributions are mistyping key names ([`githandle` instead of `githubhandle`](https://github.com/githubschool/open-enrollment-classes-introduction-to-github/pull/3429/files#diff-7ab06800da1cf47dc8114666ce54dcc1L2), for example).

brianamarie:
@swinton Thank you for clarifying. I believe that latitude and longitude are already checked as fields in `test_coordinateValidation.rb` but there's no easy way to add a check for githubhandle since it's not a number. 🙇 :shipit:

brianamarie:
@swinton I hope it's OK, I made a slight change so the build would pass by adding an error message to the `assert_true` method.

tests/test_frontmatterValidation.rb
```diff
@@ -4,18 +4,17 @@
...
+            assert_true(keys.include? "longitude")
+
             lines = File.readlines(path)
-            patterns.each_with_index do |pattern, index|
-                assert_match(pattern, lines[index])
-            end
+            assert_equal("---\n", lines[0])
+            assert_true(lines[4..-1].include? "---\n")
         end
     end
```
stoe: That looks great @swinton

brianamarie:
@swinton These tests work beautifully, thank you! My one question, is after running it locally, it's difficult to find which file is failing. Is there a way to add that information into the output so it's easier to debug?

tests/test_frontmatterValidation.rb
```diff
@@ -0,0 +1,21 @@
+require 'test/unit'
+require 'yaml'
+
+class TestFrontMatterValidation < Test::Unit::TestCase
+
+    def test_frontmattervalidation
+        Dir['./_pins/*.json'].each do |path|
+            data = YAML.load_file(path)
+
```
stoe: > Is there a way to add that information into the output so it's easier to debug?

```rb
puts "#{path}"
```

💭 s @brianamarie @swinton?

swinton: That could become quite verbose.

I wonder if we can catch the failing assertion, log the offending file, and then re-raise the assertion? I'm thinking with my Python hat on, not sure how we'd achieve that in Ruby.

_Googling_...

stoe:
```rb
assert(keys.include? "githubhandle", "githubhandle missing: #{path}")
assert(keys.include? "latitude", "latitude missing: #{path}")
assert(keys.include? "longitude", "longitude missing: #{path}")

lines = File.readlines(path)
assert_equal("---\n", lines[0], "opening --- missing: #{path}")
assert(lines[4..-1].include? "---\n", "closing --- missing: #{path}")
```

should do it !?

swinton: ca9b13531cddba399f696c50bcaf4db209f18c02

swinton: [Travis didn't like that](https://travis-ci.org/githubschool/open-enrollment-classes-introduction-to-github/builds/201584153) :smile:

It's possible specifying a `.ruby-version` will help?

swinton: 3766c7e got it :+1:

Good to :ship:?

PR #3434, opened 2017-02-03
Adding my pin

soyee:
👋 Hello @githubteacher, I've added my lat and lon to the pin file, which  closes #3434


soyee:
Fixed pull request content to include training team and <closes #my-issue-number>

PR #3436, opened 2017-02-03
LatLonjsin

DazedB74:


_pins.daze/jsin
```diff
@@ -0,0 +1,2 @@
+Your latitude.: 46.7172
```
crichID: @DazedB74 please double check the syntax as shown at https://services.github.com/on-demand/intro-to-github/08/. If you need help fixing it, check out this video: https://vimeo.com/201011288

DazedB74: Fixed it thanks.

PR #3487, opened 2017-02-05
created blake1991.json

blake1991:
added myself as a pin
@githubteacher


crichID:
@blake1991 your file is passing the tests so you can merge if you would like! It might also be interesting to add more specific coordinates! Just in case you need help editing your file, we have created a short video: https://vimeo.com/201011288

PR #3503, opened 2017-02-05
Adding Paritosh's Coordinates

paliwalparitosh:
Adding Paritosh's Coordinates @githubteacher close #3


crichID:
@paliwalparitosh your tests are passing so you can merge if you would like, but it might be more fun if you add real coordinates! If you need help making changes to your file, we have created a short video with instructions: https://vimeo.com/201011288

PR #3574, opened 2017-02-07
Adding my Coordinates 

gharam37:
Test

### DIFF
--- a/pins
+++ b/pins
@@ -1,0 +1,3 @@
+githubHandle: gharam37
+latitude: 30.0771
+longitude: 31.2859
--- a/pins
+++ b/pins
@@ -1,3 +1,3 @@
 githubHandle: gharam37
-latitude: 30.0771
+latitude: 30.0772
 longitude: 31.2859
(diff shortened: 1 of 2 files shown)


to-dy:
@gharam37 

Add the three dashes, and you should be good to go!
```
---
githubHandle: YOUR-USERNAME
latitude: YOUR_LATITUDE
longitude: YOUR_LONGITUDE
---
```

pins
```diff
@@ -0,0 +1,3 @@
+githubHandle: gharam
```
obaceski: I think you are missing the 3 dashes

pins
```diff
@@ -0,0 +1,3 @@
...
+latitude: 30
+longitude: 70
```
obaceski: again, missing 3 dashes at the end

richardrubert:
motivation and support... go you

quinmixon:
@gharam37 keep up the good work 👍

brianamarie:
Hi @gharam37, everything that's been mentioned is correct! Line one should only be the characters `---`, and the last line should be, too. Let me know if you need any help!

brianamarie:
Hi @gharam37! I went ahead and made some changes. Since it's been open for a while, I'm going to merge, too. Nice work!

PR #3596, opened 2017-02-08
Add githubstudent'spins to the map

dangvanquan:
Adding SF to the map!


_dangvanquan.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle:dangvanquan
```
crichID: @dangvanquan you will need a space after each `:`

_dangvanquan.json
```diff
@@ -0,0 +1,4 @@
...
+latitude:21.027764
+longtitude:105.834160
```
crichID: @dangvanquan you will need to add a line to the end of your file with three dashes `---`

PR #3702, opened 2017-02-12
Adding DirkStagger's coordinates

DirkStagger:
@githutteacher How did I do?

Adding DirkStagger-Edina pin to the map closes #927.


naveenchandu:
hiii add me

PR #3860, opened 2017-02-17
Create zahodinikola.json

zahodinikola:


zahodinikola:
added my city coordinates

PR #3975, opened 2017-02-20
Create OldHWGuy.json

OldHWGuy:
Adding Pin for my location

@githubteacher
closes #3975


OldHWGuy:
Committing pin to the map

PR #4052, opened 2017-02-22
Add publicdoename's pin to map

publicdoename:

This pull request closes #4052.


publicdoename:
@githubteacher Am I doing this right?
This pull request closes #4052.

Issue #4203, opened 2017-02-28
Debra Joseph Issue 1


debrajoseph:
Hello! :wave: Please add me as a collaborator to this repo.

PR #4211, opened 2017-02-28
Adding debrajoseph co-ordinates

debrajoseph:

This closes issue #4203

### DIFF
--- a/_pins/debrajoseph
+++ b/_pins/debrajoseph
@@ -1,0 +1,5 @@
+---
+githubHandle: debraj
+latitude: -33.8227
+longitude: 150.9239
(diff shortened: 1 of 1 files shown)


themooselet:
Hi Debra, could the problem be that the handle in your file is not your full username?

brianamarie:
Hi @debrajoseph! I think @themooselet is on the right track. You will need to change your file name so it ends in `.json` so the tests will pass. Let me know if you want any help!

brianamarie:
Hi @debrajoseph. I'm going to close this pull request since you've opened another with passing tests. 😄

PR #4215, opened 2017-02-28
Chris manchester

ckpilling:
hello


themooselet:
Hmm, the only difference I see between your file and mine is the "---" missing at the beginning and the end of yours. Not sure if that's the reason for the failed checks.

brianamarie:
Hi @ckpilling! @themooselet is exactly right. Try adding `---` for the first and last lines of the file. Let me know if you want any help!

PR #4259, opened 2017-03-01
Add csepanyij's pin to map

csepanyij:
Adding Budapest to the map!

### DIFF
--- a/_pins/csepanyij.json
+++ b/_pins/csepanyij.json
@@ -1,4 +1,5 @@
+---
 githubHandle:csepanyij
-latitude:47.497912
-logitude:19.040235
-
+latitude: 47.497912
+logitude: 19.040235
+---
--- a/_pins/csepanyij.json
+++ b/_pins/csepanyij.json
@@ -1,5 +1,5 @@
 ---
-githubHandle:csepanyij
+githubHandle: csepanyij
 latitude: 47.497912
 longitude: 19.040235
 ---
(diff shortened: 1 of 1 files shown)


_pins/csepanyij.json
```diff
@@ -0,0 +1,4 @@
+githubHandle:csepanyij
```
brianamarie: Try adding a line of three hyphens as line one (like this: `---`), and then have githubHandle on line 2.

brianamarie: Make sure there is a space after the colon and before `csepanyij`, and follow that pattern for latitude and longitude.

_pins/csepanyij.json
```diff
@@ -0,0 +1,4 @@
...
+latitude:47.497912
+logitude:19.040235
+
```
brianamarie: Add the same `---` on the last line, too.

_pins/csepanyij.json
csepanyij: Sorry, I forgot about these, I think I got it right now.

tuffix:
Hej @csepanyij try changing logitude to longitude! I think that should work :)

tuffix:
Hej @csepanyij,
Missed this last time but ad a space, after githubHandle: csepanyij think it should work than!
Good luck!!

crichID:
@easpaas please do not close pull requests and issues that you did not create! We know it is tempting, but we want everyone to have the opportunity to finish the project on their own. If you are interested in helping, then leaving supporting comments is :sparkles: and encouraged.

PR #4320, opened 2017-03-03
adding nationofkim coordinants

nationofkim:
--- [ ] @githubteacher  What do you think?
This pull request closes #4320.
---


_pins/nationofkim.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: nationofkim
```
brianamarie: Hi @nationofkim! To get the tests to pass, try adding a line at the beginning with just `---` and also a line at the end with `---`. Let me know if you need help!

PR #4321, opened 2017-03-03
Add ezrinjaz's pin to the map

ezrinjaz:
Adding Red Oak, OK to the map!


_pins/ezrinjaz.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: ezrinjaz
+latitude: 34.950932
+longitude: -95.080240
```
brianamarie: Hi @ezrinjaz! Try adding a line at the end with `---`. Let me know if I can help!

PR #4332, opened 2017-03-04
Add gwpjamie's pin to the map

gwpjamie:
Adding Tulsa to the map!


AShankle:
Also in Tulsa!

tuffix:
Hej @gwpjamie, think you need to direct the commit to the **_pins** directory and it should merge just fine!

gwpjamie:
@AShankle will you be at the connected Tulsa event today too?

brianamarie:
@gwpjamie I think @tuffix is right about the name of the file. Once it's `_pins/gwpjamie.json` instead of just `gwpjamie.json`, the tests should pass. Let me know if you want any help!

PR #4352, opened 2017-03-04
Dil2000 atlanta

Dil2000:
I'm new to github. Could we open a github file directly in a web browser ? 
Thanks

### DIFF
--- a/dil2000.js
+++ b/dil2000.js
@@ -1,0 +1,1 @@
+java script
--- a/dil20005
+++ b/dil20005
@@ -1,0 +1,5 @@
+---
+githubHandle: YOUR-USERNAME
+latitude: YOUR_LATITUDE
+longitude: YOUR_LONGITUDE
+---


dil2000.js
```diff
@@ -0,0 +1 @@
+java script
```
brianamarie: Try deleting this file.

dil20005
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: YOUR-USERNAME
```
brianamarie: These are just placeholders, so you'll need to replace `YOUR-USERNAME` with your actual username, so the line looks like this: `username: Dil2000`. It will be the same for latitude and longitude, too.

Dil2000: Thank you so much for your input. I really appriciate. yeah i noticed that. I was just playing with that project to get an idea about github. Is it nessessory to pass it ? I was busy creating a new file with my desktop app.

brianamarie: I'm not sure I understand. Yes, for the tests to pass, you will need to change it. BUT, if you don't care if the tests pass and feel you've gotten enough from the tutorial, you're welcome to stop now and press the grey 'close pull request' button near the bottom of the page. 😄

PR #4409, opened 2017-03-07
Adding to dah512's webpage and page directory

dah512:
what do you think, @adrianz?


dah512:
Had to reload page, got caught in loop, then had to re-create pull.  May have pushed button to many times but it locked up and now I am failing status checks

dah512:
now how do I get this to pass checks?

hollenberry:
Hey @dah512 ! I think you might have jumped into the wrong repository by mistake. It looks like you're looking for this one: https://github.com/githubschool/on-demand-github-pages

PR #4419, opened 2017-03-08
Added coordinates for Portland

easpaas:


_pins/easpaas.json
```diff
@@ -1,5 +1,5 @@
 ---
-githubHandle: easpaas
-Latitude 45.5144
-Longitude -122.6433
+githubHandle:easpaas
+Latitude:45.5144
```
easpaas: @githubteacher Do I need to put a space after the colon?

PR #4433, opened 2017-03-08
Adding IAMGABBYJ coordinates

IAMGABBYJ:


tuffix:
hej @IAMGABBYJ  make it a json file (IAMGABBYJ.json) and you should be good to go

PR #4469, opened 2017-03-09
Create I live on the road

DungHook:
There is no way I can add a pin because I don't have a permanent residence. LOL
It's very easy to make some words **bold** and other words *italic* with Markdown. You can even [link to Google!](http://google.com)


hollenberry:
Hey @DungHook ! 😆 Looks like there might not be anything to merge, then! But good job creating the PR! 

If you want to create a file that reveals a location that's important to you, feel free to do that! Otherwise, it might be good to close this PR out. 👍

snaduvinamani:
Good Job !

PR #4490, opened 2017-03-11
Pin:  Latitude and Longitude

fusionrace:
Added Latitude and Longitude for username fusionrace


fusionrace:
@githubteacher closes #1

PR #4578, opened 2017-03-15
Adding SemperFratres' coordinates

SemperFratres:


_pins/SemperFratres.json
```diff
@@ -0,0 +1,3 @@
+githubHandle:SemperFratres
```
crichID: @SemperFratres you will need to add a line at the beginning and end of the file that just contains three dashes `---`.  You will also need a **space** after each `:`. If you need any help making the changes, I created a quick video to show you how: https://vimeo.com/201011288

PR #4595, opened 2017-03-16
Add ijhajanellejacob's pin to the map

ijhajanellejacob:
Adding ijhajanellejacob to the map!

### DIFF
--- a/_pins/ijhajanellejacob.json
+++ b/_pins/ijhajanellejacob.json
@@ -1,0 +1,5 @@
+---
+ijhajanellejacob: IjhaJanelleJacob
+latitude: 40.668949
+longitude: -73.7815625
+---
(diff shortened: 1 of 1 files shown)


M-W-C:
Hi @ijhajanellejacob 

From what I can see in the Files changed section, you need to make a change in the syntax used.

Everything looks fine at a glance except for line 2 which reads:

ijhajanellejacob: IjhaJanelleJacob

**Solution:**

Give the following syntax a try. Just replace the bold-hyphenated text with the appropriate information. 
____________________________
githubHandle: YOUR-USERNAME
latitude: YOUR_LATITUDE
longitude: YOUR_LONGITUDE
____________________________
Overall things look fine and should work after this.

PR #4646, opened 2017-03-18
adds username label in map

shajanjp:
Github handle will be shown as label for each pin

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,6 +33,7 @@
           var latLng = new google.maps.LatLng(coords[1],coords[0]);
           var marker = new google.maps.Marker({
             position: latLng,
+            label: results.features[i].properties.username,
             map: map
           });
         }


shajanjp:
@githubrobot @crichID is this ok ?

MixMasterT:
@shajanjp nice work! You are making great progress.

hectorsector:
Thanks for the contribution, @shajanjp! I'll ping other members of @githubschool/training-team so we can get some 👀 and code review kicked off!

jarryh:
@shajanjp  Hey cool 👍

hectorsector:
@stoe @shajanjp I don't think this actually changes the display on the published site -- just on the internal rendering of this repo. Therefore, I think in order to display this we'll need an accompanying contribution to the [corresponding `js` file on `training-kit`](https://github.com/github/training-kit/blob/master/assets/js/mapsAPI.js).

stoe:
> Therefore, I think in order to display this we'll need an accompanying contribution to the corresponding `js` file on `training-kit`.

@hectorsector @shajanjp , yes.
We need to add the `label` attribute also in https://github.com/github/training-kit/blob/master/assets/js/mapsAPI.js#L47-L61 for it to show up. Right now it only has a `title` attribute that will display the user's name on hover only.

shajanjp:
@stoe @hectorsector It will be so awesome if we can replicate the same map design in this page https://services.github.com/on-demand/intro-to-github/ :relaxed:

PR #4658, opened 2017-03-18
Add my pin - Damob28 Ireland

DamoB28:
Add to the map


_pins/DamoB28.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHanled: DamoB28
```
emesterhazy: I think there is a typo here. Should read 'githubHandle:'

PR #4717, opened 2017-03-21
Add amitbasuri123's pin to the map.

amitbasuri123:


a-a-ron:
Hey @amitbasuri123 , make sure that your `.json` file is created in the `_pins` directory not `.pins`. Make that file path change and your builds should pass. 👍

hectorsector:
@amitbasuri123 you're almost there! Just follow @a-a-ron's advice and move your pin into the `_pins` directory. [Here's a video](https://vimeo.com/201011288) that demonstrates how to fix that issue, along with a few others.

PR #4749, opened 2017-03-22
update coordinates in gem.zhang.json

zppet:


zppet:
we are done!
Thanks.
Let's go check the map and see if the icon has been moved.

PR #4857, opened 2017-03-24
Adding Danderhall to the map

paralerouss:
paralerouss adding a ping for Danderhall to the map


mburghelea:
Looking good

nbaillie:
nice work as expected :1st_place_medal:

samyool:
Sensational. You are the best

PR #4875, opened 2017-03-25
adding Chiriboga's pin to the map

chiriboga:


chiriboga:
adding myself to the map

PR #4904, opened 2017-03-26
Joelgodin venice

JoelGodin:
pull


JoelGodin:
wth

JoelGodin:
had to add three zeros to lat long

PR #5047, opened 2017-03-30
Add cheperth pin to the map

cheperth:


ofirgut007:
@cheperth this looks great 🎉  📢

PR #5194, opened 2017-04-04
Add my pin to the map!

AntonLedorub:
Adding L'Hospitalet to the map!


shajanjp:
Your pin should be in this format.
```
---
githubHandle: <YOUR_USER_NAME>
latitude: <LATITUDE>
longitude: <LONGITUDE>
---
```

you forgot to add `---` in the last line.

PR #5277, opened 2017-04-06
Add btskloud pin to the map

btskloud:
add pin to map


crichID:
@btskloud it looks like you have two pull requests open. I am going to go ahead and close this one since the other looks ready to merge :+1:

PR #5439, opened 2017-04-11
Create pin for @ani6gup

ani6gup:
Does it look ready to go?


_pins/ani6gup.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: ani6gup
+Your Latitude: 33.4357
```
joenash: should just be "latitude" and "longitude", that should fix it! :)

ani6gup: whoops, thank you!

PR #5668, opened 2017-04-19
Add thutcheson-stoneriver's pin to the map

thutcheson-stoneriver:
Adding Orlando, FL to the map

### DIFF
--- a/pins/thutcheson-stoneriver.json
+++ b/pins/thutcheson-stoneriver.json
@@ -1,0 +1,5 @@
+---
+githubhandle: thutcheson-stoneriver
+latitude: 28.589
+longitude: -81.206
+---


crichID:
@thutcheson-stoneriver it looks like you closed your pull request instead of merging it. When you close the pull request, it is like saying "never mind - I don't want to make this change". However, if you want your pin to appear on the map, you will need to merge it. 

If you would like to continue the activity, please feel free to reopen the pull request and press the big green button.

thutcheson-stoneriver:
I realized two things when the pull request failed:
1. The githubhandle needed to have a capital 'H' in handle
2. I put the file in /pins instead of /_pins

I created a new file in the correct directory, I deleted the incorrect file and closed that pull request.

My guess is that there was a better way to solve this but that was my initial reaction.  What should I have done?

crichID:
@thutcheson-stoneriver your way works, but you can also edit directly in the pull request. Just so you know for next time, here is a quick video to show you how: https://vimeo.com/201011288

thutcheson-stoneriver:
@crichID Oh, cool!  That video was very helpful.  I did not know that changing the directory was that easy.  Thanks!

PR #5706, opened 2017-04-20
my postion

baruchMo:
cordinate
@githubteacher -what next?
closes #5705

### DIFF
--- a/_pins/baruchMo.json
+++ b/_pins/baruchMo.json
@@ -1,3 +1,5 @@
+---
 githubHandles: baruchMo
 latitude: 32.3309
 longitude: 34.8599
+---
--- a/_pins/baruchMo.json
+++ b/_pins/baruchMo.json
@@ -1,3 +1,5 @@
----githubHandles:baruchMo
+---
+githubHandle:baruchMo
 latitude:32.3309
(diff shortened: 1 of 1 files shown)


crichID:
@baruchMo your pull request is very close. All you need to do is add three dashes `---` to the beginning and end of the file. We are using yaml front matter to create a jekyll collection and these dashes are what indicates the text you entered is frontmatter :wink: If you need any help, this video has instructions for fixing common issues: https://vimeo.com/201011288

_pins/baruchMo.json
```diff
@@ -0,0 +1,3 @@
+---githubHandles:baruchMo
```
crichID: @baruchMo you will want to add a line break after the `---` also, it should be `githubHandle` instead of `githubHandles` and you will need a space after the `:` on each line 😄

_pins/baruchMo.json
```diff
@@ -0,0 +1,3 @@
...
+latitude:32.3309
+longitude:34.8599---
```
crichID: @baruchMo you will want to do a hard return after the last 9 so the `---` is on a line by itself!

PR #5821, opened 2017-04-24
add jonathansc83's pin to the map

jonathansc83:
adding location


jonathansc83:
adding location

brianamarie:
Hi @jonathansc83! Make sure that your file has the line characters (`---`) at the bottom of the file. Also, make sure the name ends in `.json`, with no other characters at the end. 

Let me know if you want or need help making those changes so the test will pass. 😄

PR #5835, opened 2017-04-24
Create Righteous_Jams

VinnieLu:
My coordinates are


brianamarie:
Hi @VinnieLu! I think the tests are failing because of the filename. It needs to end in `.json`. Let me know if you want or need any help!

PR #5837, opened 2017-04-24
Create _pins\stormrider41.json

stormrider41:
Adding to map


brianamarie:
Hi @stormrider41! A few changes that may need to happen for the tests to pass:
-The name is close, but I think the `/` is going the wrong direction. Try changing this.
- Make sure that `githubHandle` has the H capitalized 

Let me know if you want or need any help. 😄

PR #5940, opened 2017-04-27
Create ursreenu.json

ursreenu:
Add my Location

@githubteacher 
 How is it closes #3.


brianamarie:
Hi @ursreenu! I've renamed the file from `pin` to `pins` so the tests will pass. Let me know if you want or need any help! 😄

PR #5944, opened 2017-04-27
Add githubstudent-melbourne-au's pin to map

martinhalford:
Adding Melbourne AU to the map!


_pins/githubstudent-melbourne-au.json
```diff
@@ -0,0 +1,5 @@
+---
+githandle: githubstudent-melbourne-au
```
leera01: hello martin, kindly change the value to githubHandle so that the build will pass.

PR #5946, opened 2017-04-27
Adding the default file

atinmalik:
@githubteacher


brianamarie:
Hi @atinmalik! I've made some changes to the file, only adding `---` to the beginning and end. I didn't change any of the other content. Hopefully, the tests will pass now. Let me know if you need or want any help. 😄

Issue #6002, opened 2017-04-29
Please add me, Orlando, FL


wfrank1155:
Hello! :wave: Please add me as a collaborator to this repo. 

28.5383° N, 81.3792° W

githubteacher:
:tada: @wfrank1155 you are now a repository collaborator. :balloon:

githubteacher:
It looks like you have already merged your pull request or you stopped working on the activity. I'm going to go ahead and close this issue to keep the repository tidy.

PR #6010, opened 2017-04-29
Adding wfrank1155's coordinates

wfrank1155:
Hello @githubteacher 

Not sure what issue number should be #6010 ?????


SandMV:
Hello, @wfrank1155! Your issue number is #6002
Well done and good luck!

_pins/wjfrank1155.json
```diff
@@ -0,0 +1,6 @@
+...
```
brianamarie: Hi @wfrank1155! It looks like you're using `.` instead of `-`. Try replacing it so the first and last line has `---`, and see if that works.

wfrank1155: Thanks @brianamarie .  Indeed that was the problem. I should've just copy/pasted :-)

wfrank1155:
This was an excellent tutorial

PR #6024, opened 2017-04-29
Githubstudent plano

lakhani9:


brianamarie:
@lakhani9, @ElaCai is exactly right! Please delete two of these files, and edit the name of one so it ends in `.json`. Let me know if you need or want any help!

PR #6080, opened 2017-05-02
Revert "Add ck's coordinates"

chloekrueger:
closes #6078 @githubteacher


chloekrueger:
What do you think @githubteacher? Please add me.

PR #6145, opened 2017-05-03
Add mhowardmo's pin to the map

mhowardmo:
File with lat/long of KHS to create pin on map


mhowardmo:
Adding mhowardmo's pin to the map

mhowardmo:
Pin should be added!

PR #6161, opened 2017-05-04
Add NoraMcGregor's pin to the map

NoraMcGregor:
:+1:

### DIFF
--- a/_pins/NoraMcGregor.json
+++ b/_pins/NoraMcGregor.json
@@ -1,0 +1,4 @@
+---
+githubHandle: NoraMcGregor
+latitude: 51.529985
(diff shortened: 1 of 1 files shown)


_pins/NoraMcGregor.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: NoraMcGregor
+latitude: 51.529985
+longitude: -0.127014
```
hectorsector: @NoraMcGregor you just need the closing `---` at the bottom of the code.

NoraMcGregor: Ahh thanks, missed that in the video! Will try again!

PR #6367, opened 2017-05-10
Add jellis002's pin to the map

jellis002:
Adding newington to the map.


hollenberry:
@jellis002 You were so close here! I don't know if you reverted on purpose, but if you change your mind and want to try again your file looked perfect. All you needed was to do a rename of the file. 

Please let us know if you want to try again or need help!

PR #6394, opened 2017-05-10
Add VOffoma's coordinates

VOffoma:
@githubteacher What do you think?

This pull request closes #6390

PR #6395, opened 2017-05-10
Create my pin on map

lenfien:
Ad SF to the map!

close #6394


pachoyan:
@lenfien Go!
:loudspeaker: :tada:.

PR #6426, opened 2017-05-11
Create -pins-pbeene

pbeene:
Adding my location as a pin
@githubteacher What do you thing?
closes #6426


-pins-pbeene
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: pbeene
+latitude: 39.9469
+longitude: -105.2142
+--- adding my pins to github handle
```
crichID: @pbeene please remove the extra content after the three dashes. The last line should only be `---`.

jazmynejack:
if you remove "adding my pins to github handle" from the end of the file, your checks will pass

PR #6430, opened 2017-05-11
Create SARAVKOCH.json

saravkoch:
Add a new pin file for map

### DIFF
--- a/_pins/SARAVKOCH.json
+++ b/_pins/SARAVKOCH.json
@@ -1,0 +1,3 @@
+githubHandle: SARAVKOCH
+latitude: 33.440
+longitude: -112.071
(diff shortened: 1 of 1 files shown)


_pins/SARAVKOCH.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: SARAVKOCH
```
crichID: @saravkoch please add three dashes `---` on the first and last line of the file so it looks like this:

```
---
githubHandle: your-handle
latitude: your-latitude
longitude: your-longitude
---
```

saravkoch: Added dashes and committed

kenners2082:
Good job 👍  @saravkoch

PR #6431, opened 2017-05-11
Create jazmynejack.json

jazmynejack:
look at my new changes, let me know what you think ?

### DIFF
--- a/_pins/jazmynejack.json
+++ b/_pins/jazmynejack.json
@@ -1,3 +1,5 @@
+---
 githubHandle: jazmyne
 latitude: 39.9469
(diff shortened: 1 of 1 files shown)


_pins/jazmynejack.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: jazmyne
```
crichID: @jazmynejack please add three dashes `---` on the first and last line of the file so it looks like this:

```
---
githubHandle: your-handle
latitude: your-latitude
longitude: your-longitude
---
```

PR #6468, opened 2017-05-12
Add ripzdev's pin to the map

ripzdev:
Add Quito to the map!


beardofedu:
@ripzdev please change `.jason` to `.json` and the test should pass. 👍

PR #6484, opened 2017-05-12
Adding my pin to the map - mkeymont

mkeymont:
Adding my location data for the map.

@githubteacher helped me with this, so if it's wrong, it's likely their fault!!!


mkeymont:
Closes #6484

Issue #940, opened 2016-09-29
San Francisco


alrightjones:
## Adding San Francisco to the map

Hi everyone! This issue is to track the addition of San Francisco to the map, which can be found at:
- Longitude: 37.775928
- Latitude:  -122.444024

sainathdevi:
i am adding

smauvais:
open

githubstudent:
❤️ ! I'll be adding my pin as well.

PR #6487, opened 2017-05-12

dashzeveg:
Adding Ulaanbaatar to the map!

closes #940


beardofedu:
@dashzeveg your code should look something like this:
```
---
githubHandle: githubstudent
latitude: 47.886399
longitude: 106.905744
---
```
So, please delete the `...` on line 1 and replace it with `---` and add `---` to line 5.

PR #6488, opened 2017-05-12
Adding my pin to the map

dashzeveg:
Adding Ulaanbaatar to the map!

closes #940

### DIFF
--- a/_pins/dashzeveg.json
+++ b/_pins/dashzeveg.json
@@ -1,0 +1,4 @@
+...
+githubHandle: dashzeveg
+latitude: 47.886399
+longitude: 106.905744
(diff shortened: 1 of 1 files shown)


DanayaRChandler:
@dashzeveg Hiya! This looks good except I believe lines 1 and 5 should have three dashes (---) and you don't need line 6 at all! Good luck! 🤞

beardofedu:
@dashzeveg @DanayaRChandler is correct you need to change the `...` to `---` in your file to get this to pass the test.

GingerJava:
@dashzeveg I agree with  @DanayaRChandler lines 1 and 5 need 3 of these: --- instead of ... other than that, YAY! 👍 🎉

PR #6524, opened 2017-05-14
Add my handle to the map

djhernanv:
Test from Tutorial!


hollenberry:
Hey @djhernanv ! Looks like someone closed your PR when they shouldn't have! 

I looked into why your build wasn't passing and found that you ended up creating two files. I deleted the extra one that was named incorrectly, and the second is so close! 

Instead of `githubhandle`, it should be `githubHandle`. Please let us know if you have any questions! You're almost there 👍

djhernanv:
Hey @hollenberry! Thanks for the deletion. The change from `githubhandle` to `githubHandle` should be on the .json file? I did so, but the check still fails..
Do you see my mistake?
Thanks for the help!

hollenberry:
Ah, @djhernanv ! So close!

There's a minor typo in there as well.

Should read `githubHandle` instead of `gibthubHandle` <-- extra b

djhernanv:
Awesome. It worked.
Thanks!

PR #6555, opened 2017-05-15
Please add githubstudent's pin to the map

just-Meltem:
Adding TR to the map

Closes #950

### DIFF
--- a/_pins/github student.json
+++ b/_pins/github student.json
@@ -1,0 +1,3 @@
+githubHandle:  githubstudent
+latitudute: 38.614034
+longitude: 27.429562
--- a/_pins/github student.json
+++ b/_pins/github student.json
@@ -1,3 +1,0 @@
-githubHandle:  githubstudent
-latitudute: 38.614034
-longitude: 27.429562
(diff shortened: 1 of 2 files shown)


milenkobrkovic:
@just-Meltem  This looks great 🎉

_pins/github student.json
```diff
@@ -0,0 +1,3 @@
+githubHandle:  githubstudent
```
crichID: @just-Meltem please replace `githubstudent` with your username and check that there is only 1 space between the `:` and your username.

crichID: You will need to add a `---` on the first and last line of your file so it looks like this:

```
---
githubHandle: just-Meltem
latitude: 38.614034
longitude: 27.429562
---
```

hollenberry:
Hey @just-Meltem  ! This is good to merge! When you're ready, go ahead and click that big green `Merge Pull Request` button! If you don't do it by the milestone date, we'll go ahead and run that for you in the background.

Let us know if you need any help or have questions! 🎉

PR #6586, opened 2017-05-16
Add githubdatascientist's pin to the map

shahabkarimi:


shahabkarimi:
Add Carlsbad to the map!

closes #940

PR #6592, opened 2017-05-16
Create Juell1635.json

Juell1635:
Longitude and Lat for Pin location


Juell1635:
I have updated the pin information

It will be different from @githubteacher information 

closing#6592

Juell1635:
I have updated the pin information
It will be different from @githubteacher information
**closing#6592**

PR #6723, opened 2017-05-18
Add idike's pin to the map

idike:
Add SJ to map


SWaltersdorff:
@idike This looks good to me. :star2:

PR #6770, opened 2017-05-20
Add shhhhmel's pin to the map!

shhhmel:


shhhmel:
I have "All checks have failed" all time.

PR #6796, opened 2017-05-22
Create cgbarfield.json

cgbarfield:
@githubteacher
close #6796

### DIFF
--- a/_pins/cgbarfield.json
+++ b/_pins/cgbarfield.json
@@ -1,3 +1,5 @@
+---
 githubHandle: cgbarfield
 latitude: 35.8601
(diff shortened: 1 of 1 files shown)


_pins/cgbarfield.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: cgbarfield
```
crichID: @cgbarfield you are close, you just need to add three dashes on the first and last line `---` so the file looks like this:

``` 
---
githubHandle: cgbarfield
latitude: 35.8601
longitude: -78.8326
---
```

PR #6847, opened 2017-05-24
Add WrayJonathanM's pin to the map please!

WrayJonathanM:
Adding DFKI to the map if it's not already there

### DIFF
--- a/_pins/WrayJonathanM2.json
+++ b/_pins/WrayJonathanM2.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: WrayJonathanM
-Latitude: 49.429580
-Longitude: 7.751369
+latitude: 49.429580
+longitude: 7.751369
 ---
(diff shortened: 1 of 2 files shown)


crichID:
@WrayJonathanM I see you fixed the spelling mistake in the second file, but your tests will continue to fail because the first file is still part of your branch. I am going to delete the first file for you since it is a little tricky. Then the tests should pass.

crichID:
@WrayJonathanM I deleted the file so the tests "should" pass now. In case you need to know in the future, here is a video that shows how you can access the same file and make some of the common corrections: https://vimeo.com/201011288

_pins/WrayJonathanM2.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: WrayJonathanM
+Latitude: 49.429580
```
crichID: @WrayJonathanM you will need to make Latitude and Longitude both a lowercase l in order for your tests to pass!

WrayJonathanM: thanks!

brianamarie:
@WrayJonathanM I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #6850, opened 2017-05-24
add my location please Jetti1 berlin

Jetti1:
Hi @githubteacher 

please add me to the map. I created 2 files..the first one I forgot line1 and line5 with ---

### DIFF
--- a/_pins/Jetti1.json
+++ b/_pins/Jetti1.json
@@ -1,3 +1,0 @@
-githubHandle: Jetti1
-latitude: 52.5033
-longitude: 13.329
--- a/_pins/Jetti1.json
+++ b/_pins/Jetti1.json
@@ -1,0 +1,3 @@
+githubHandle: Jetti1
+latitude: 52.5033
+longitude: 13.329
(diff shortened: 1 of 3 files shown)


crichID:
@Jetti1 it looks like your last file is correct, but the checks will continue to fail as long as the two incorrect files are on your branch. We created this quick video to show you how to correct the files on your branch: https://vimeo.com/201011288

In the meantime, I will delete the extra files since this process can be confusing :wink:

Jetti1:
@crichID cool, thanks a lot. in one file i missed the --- at the beginning and at the end. But was was the second failure?

crichID:
> But was was the second failure?

The second file was only failing because the first file was still a part of the branch. The tests check everything on your branch since that is what you are proposing to merge in to master. To prevent this, you could just edit the first file (check out that video above).

Jetti1:
@crichID yep, i saw it. Thanks a lot. Instead of editing the first file i created 2 new ones and tried somehow to delete the first one. But now I know how to editing a file.

PR #6865, opened 2017-05-24
Adding stephaniewatler coordinates

stephaniewatler:


_pins/stephanewatler.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: STEPHANIEWATLER
+latitude: 19.2667
+longditude: -81.3
```
crichID: @stephaniewatler it looks like an extra letter snuck in here. It should be spelled `longitude`. If you fix that, your tests should pass and you will be able to merge. If you need any help, please check out this video on fixing common problems: https://vimeo.com/201011288

brianamarie:
@stephaniewatler  I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #6867, opened 2017-05-25
add githubstudentwb's pin to the map

BikashPandey17:
Adding WB to tthe map!

### DIFF
--- a/_pins/githubstudentwb.json
+++ b/_pins/githubstudentwb.json
@@ -1,0 +1,4 @@
+---
+githubhandle: githubstudentwb
+latitude: 22.748331
+longitude: 88.338505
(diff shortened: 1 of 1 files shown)


_pins/githubstudentwb.json
```diff
@@ -0,0 +1,4 @@
+---
+githubhandle: githubstudentwb
```
crichID: @BikashPandey17 you will want to change this to your username so it will appear on the map with your avatar!

_pins/githubstudentwb.json
```diff
@@ -0,0 +1,4 @@
...
+latitude: 22.748331
+longitude: 88.338505
```
crichID: @BikashPandey17 please add three dashes `---` to the last line of this file. It should look just like the first line.

brianamarie:
@BikashPandey17  I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #6876, opened 2017-05-25
add githubstudnet pin to the map

Nazmul4240:
Add me


schowalk:
@Nazmul4240 Nice work!!

_pins/nazmul.json
```diff
@@ -0,0 +1,5 @@
+...
```
crichID: @Nazmul4240 you will need to change these dots to dashes `---` on both the first and last line.

_pins/nazmul.json
```diff
@@ -0,0 +1,5 @@
...
+githubhandle: nazmul
+Latitude:24.368713
```
crichID: @Nazmul4240 please make this a lowercase `l` and add a space after the `:`. This also applies to the longitude line!

PR #6903, opened 2017-05-26
add github-henryho

ecshenryho:


_pins/github-henryho
```diff
@@ -0,0 +1,5 @@
+---
+githubhandle:ecshenryho
```
crichID: @ecshenryho you will need to have a space between the `:` and your username!

_pins/github-henryho
```diff
@@ -0,0 +1,5 @@
...
+Latitude: 33.825464
```
crichID: @ecshenryho both Latitude and Longitude should start with a lowercase l.

krudge:
Nice!

ChasSykes:
Hi ecshenryho,

I think another roadblock to successful build might be that the filename is supposed to be your handle and the extension ".json". Try naming your file "ecshenryho.json", in addition to whatever other recommendations were made before mine.

brianamarie:
@ecshenryho  I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #6912, opened 2017-05-26
Add Meeett's pin to Map

Meeett:
Adding my Pin


Meeett:
Add my pin to the map.

PR #6922, opened 2017-05-26
Create krudge.json

krudge:
Change this @githubteacher

### DIFF
--- a/_pins/krudge.json
+++ b/_pins/krudge.json
@@ -1,0 +1,4 @@
+---
+githubHandle: krudge
+latitude: 40.4344
+longitude: -80.0248


krudge:
@githubteacher  what's wrong?

_pins/krudge.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: krudge
```
crichID: @krudge you are close - you just need to add three dashes at the beginning and end of your file `---`, so your file would look like this:

```
---
githubHandle: krudge
latitude: 40.4344
longitude: -80.0248
---
```

PR #7082, opened 2017-06-01
Adding rmwegle coordinates

rmwegle:

This closes issue #7081


rmwegle:
Merging rmwegle location coordinates

PR #7093, opened 2017-06-01
Add my current location, though it's a secret

vkrauze:
Yeah, I'm giving out my coordinates. Please don't it for a nuclear missle target practice! closes #7093


vkrauze:
Good work, buddy!

PR #7106, opened 2017-06-01
Add db's pin to the map (attempt #2)

deepak-behera:
Adding my location to the map.


brianamarie:
@deepak-behera 👋 Hello! I  think the reason the tests are failing is because of the name of the file. Try renaming the directory to `/_pins/` instead of `/..pins/` and see if that helps. Let me know if you need or want help. 😄

PR #7131, opened 2017-06-02
Add bacoder28 pin to the map

badcoder28:
Adding PH student to the map!

### DIFF
--- a/_pins/badcoder28.json
+++ b/_pins/badcoder28.json
@@ -1,0 +1,4 @@
+---
+githubHandle: badcoder28
+latitude: 10.636378
+Longitude: 122.955381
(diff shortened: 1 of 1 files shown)


badcoder28:
ops! and why?

brianamarie:
Hi @badcoder28! A few things that might be causing the tests to fail: 
- try `latitude` instead of `Latitude`, this is case sensitive 
- Make sure there are three lines `---` at the bottom of your file as well

Let me know if you need or want any help 😄

badcoder28:
@brianamarie hello again, i followe your instructions and recheck the file but its still failing, any idea? thanks!

PR #7201, opened 2017-06-05
Pávlos Kodikós: Brooklyn, NY

PavlosKodikos:
---
githubhandle: PavlosKodikos
latitute: 40.691285
longitude: -73.944994
---


PavlosKodikos:
I wonder what happens if the branch is left ... ie, if it isn't deleted?  other than space on the repository, is there any other effect?  curiously, Pavlos

PR #7228, opened 2017-06-06
Add Jay C's pin to folder

jaychiarella:


_pins/jaychiarella
```diff
@@ -0,0 +1,4 @@
+--- 
+gitHubHandle: jaychiarella
+latitude: 43.6532° N
+longitude: 79.3832° W
```
hectorsector: Great work so far! You'll need to make sure that your code matches exactly the template in the class, including `---` before and after the code:
```
---
githubHandle: YOUR-USERNAME
latitude: YOUR_LATITUDE
longitude: YOUR_LONGITUDE
---
```

In addition, ensure that the file is name `jaychiarella.json`.

stevenwang0618:
Delete `° N` and `° W` at the end of latitude and longitude. And add `---` after the code.

PR #7238, opened 2017-06-06
Add rwalker707's pin to the map

rwalker707:
Adding Carbondale to map


pins/rwalker7072.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: rwalker707 
+latitude: 39.400739
+longitude: -107.208877
+---
```
hectorsector: @rwalker707 you can remove this file altogether

pins/rwalker707.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: rwalker707
+latitude: 39.400739
+longitude: -107.208877
+---
```
hectorsector: @rwalker707 for this file, just ensure that the directory name starts with an underscore: `_pins/rwalker707.json` -- just holler if you want some help with it!

PR #7264, opened 2017-06-07
Aytaylor charlottesville

Aytaylor:
Add this pin

### DIFF
--- a/_pins/aytaylor.json
+++ b/_pins/aytaylor.json
@@ -1,5 +1,5 @@
 ---
-hithubhandl: aytaylor
+githubHandle: aytaylor
 latitude: 38.907192
 longitude: -77.036871
 ---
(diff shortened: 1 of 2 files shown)


_pins/byassine93.yaml
```diff
@@ -0,0 +1,5 @@
+ ---
+ githubHandle: byassine93
+ pageUrl: byassine93.github.io
+ timestamp: TODAYS-DATE-2017-06-05
+ ---
```
hectorsector: @Aytaylor I'd remove this file since it looks like it belongs to another student.

_pins/aytaylor.json
```diff
@@ -0,0 +1,5 @@
+---
+hithubhandl: aytaylor
```
hectorsector: make sure this says `githubHandle` exactly.

MaruthiMohanReddyPutha:
@Aytaylor I think you can do this! Way to go!

PR #7316, opened 2017-06-08
Add gavilan's pin to the map

gavilan:
Add Gavilan to the map!

### DIFF
--- a/_pins/gavilans.json
+++ b/_pins/gavilans.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: gavilan
-latitud: 9.927701
-longitud: -84.001926
+latitude: 9.927701
+longitude: -84.001926
 ---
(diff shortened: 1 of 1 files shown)


_pins/gavilans.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: gavilan
+latitud: 9.927701
```
stevenwang0618: You miss an `e` at the end of this word.

_pins/gavilans.json
```diff
@@ -0,0 +1,5 @@
...
+longitud: -84.001926
```
stevenwang0618: You also miss an `e` here.

gavilan: oops, sorry, let me fix and commit again

PR #7401, opened 2017-06-12
Add pin for socorsan

socorsan:
Add pin for socorsan on the map


socorsan:
@githubteacher closes #7401

PR #7442, opened 2017-06-12
Mirmohsinahmed ka

mirmohsinahmed:
Adding KA to map


shajanjp:
@mirmohsinahmed almost there. The json file is case sensitive. Change `githubhandle` to  `githubHandle`. Fix other mistakes too.
Your pin should be in this format.
```
---
githubHandle: <YOUR_USER_NAME>
latitude: <LATITUDE>
longitude: <LONGITUDE>
---
```

PR #7484, opened 2017-06-13
Adding sogito12 coordinates

Sogito12:
Adding the city of Butler to the map!

:fist_right: :fist_left:


AlbertoMera:
Coolio, @Sogito12

PR #7533, opened 2017-06-14
Kobe's Atlanta coordinates

kobe1k:
adding Atlanta to the map


Bobelon:
Your's file name is kobe1k, but it should be kobe1k.**json**. Then all should be OK

beardofedu:
@kobe1k I modified your file name and add spaces, this should be good to merge once the test passes.

nth-block:
@kobe1k , :tada: you are all set...

PR #7562, opened 2017-06-15
add giantthinker to the map

giantthinker:
adding Foumban to the map!


beardofedu:
@giantthinker it looks like you just need to change `githubstudent: giantthinker` to `githubHandle: giantthinker` and it should pass the test and then you can merge the Pull Request.

giantthinker:
I have changed. How can I rebuild the project?

PR #7573, opened 2017-06-15
Add SeanCK 's pin to the map

SeanCK:
Adding SF to the map!!


hollenberry:
@SeanCK it looks like you accidentally closed this, but you should be able to merge without issues! Did you meant to close instead?

PR #7586, opened 2017-06-15
Add hiteshpatelgenco pin to the map

hiteshpatelgenco:
Adding hiteshpatelgenco to the map


beardofedu:
@hiteshpatelgenco I believe you need to provide additional digits in your `latitude` and `longitude` before your test will pass.

beardofedu:
@hiteshpatelgenco this is good to merge. 👍

PR #7627, opened 2017-06-17
Adding my pin to the map

ikhall:
Entering this to add my pin to the map. Will continue learning to use GitHub!


ikhall:
@githubteacher closes

PR #7754, opened 2017-06-20
Update anitconsultantllc.json

anitconsultantllc:
Updated missing Longitude (again, it had two dashes instead of one).


anitconsultantllc:
@githubteacher Why is there no pin currently added to this location:  latitude: 37.732726, longitude: -87.182171?

PR #7787, opened 2017-06-21
Pin KaPrz 

KaPrz:
This pull request closes #7787.


KaPrz:
This pull request closes #7787.

PR #7822, opened 2017-06-22
add githubstudents's pin to the map

alvindivina08:
Adding Chi to the map!


hollenberry:
Almost there @alvindivina08 ! I deleted all of your extra files (the process isn't super straightforward) and left one for you to edit.

Looks like you just need to rename your file to say `_pins_` instead of `_pin`.  

brianamarie:
@alvindivina08 I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #7835, opened 2017-06-22
Add DChickJr's Pin to the map

DChickJr:
adding CHI to the map!

closes #7835

### DIFF
--- a/_pins/DChickJr
+++ b/_pins/DChickJr
@@ -1,0 +1,1 @@
+I need help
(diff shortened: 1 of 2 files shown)


hollenberry:
Almost there @DChickJr  ! 

Looks like you just need to add three dashes below your code, like in the first line. 

brianamarie:
@DChickJr  I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #7845, opened 2017-06-22
Add boss1son pin to the map

boss1son:
Adding Brockton to the map!


boss1son:
Not sure why checks are failing

hollenberry:
Almost there @boss1son  ! 

Looks like you just need to add three dashes below your code, like so: `---`. 

PR #7895, opened 2017-06-24
Add location data to GitHubTeacher

KirinaK:
I don't know what I do and how it works.....
@githubteacher what do you think? How do I cope with learning?


shajanjp:
@KirinaK  Looks great ! Now press the big green **Merge pull request** button below and your coordinates will be added to the map.

PR #7967, opened 2017-06-27
jainshubham1506.json

jainshubham1506:


jainshubham1506.json
```diff
@@ -0,0 +1,4 @@
+---
+username:jainshubham1506
```
crichID: @jainshubham1506 you will need to change this to `githubHandle:` instead of `username:`

crichID: @jainshubham1506 you will need to add a space after each `:`.

jainshubham1506.json
```diff
@@ -0,0 +1,4 @@
...
+latitude:12.916517
+longitude:79.132499
```
crichID: @jainshubham1506 please add `---` as the last line of your file.

brianamarie:
@jainshubham1506  I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #8097, opened 2017-06-30
Add glezsotelo's pin to the map

GlezSotelo:
Adding Guadalajara to the map


ksasi:
@GlezSotelo This looks great 🎉

PR #8105, opened 2017-06-30
Adding my pin -  yes it meeee the very same elmismisimo!!!

elmismisimo:
Im adding my pin to the map, or thats what I hope im doing :D


davidnhz:
👍

PR #8269, opened 2017-07-03
Adding iam2nithin coordinates

iam2nithin:
@iam2nithin how is ur loaction? Happy !!


_pins/iam2nihin.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: iam2nithin
```
iam2nithin: wow Great

PR #8283, opened 2017-07-04
Revert "Adding pin to the map"

shaloo:
Reverts githubschool/open-enrollment-classes-introduction-to-github#8282


shaloo:
closes #8282

PR #8284, opened 2017-07-04
Shaloo pune merge request

shaloo:
Single file add shaloo.json for the pin with coordinate location


shaloo:
@githubteacher Thanks.

PR #8401, opened 2017-07-08
Add paand's pin to the map

paand:


paand:
My commit doesn't seem to pass the tests, so I cannot merge. How do I resolve this? @githubteacher

PR #8450, opened 2017-07-11
add githubStudent pin to the map

El-sakka:
Adding Eg to the map


beardofedu:
@El-sakka you need to create the file in the `_pins` folder and have `---` on a line after the code. Your tests should pass then.

PR #8473, opened 2017-07-11
Add Theoval23Blumenkraft's pin to the map

Theoval23Blumenkraft:


beardofedu:
@Theoval23Blumenkraft you need to end your file with the `.json` extension not `.NO`. Also, `githubhandle` needs to be written like: `githubHandle`. The tests should pass after you make those changes.

PR #8544, opened 2017-07-13
commit lat longitude

nanomanon56:
@githubteacher  telling people about the burbs 

closes #1


beardofedu:
@nanomanon56 the contents of your file don't match what the instructions from the class tell you to use, so the test is failing. If you modify the contents of the file, it will pass the test.

nanomanon56:
@githubteacher Hey thanks, i think i might of missed something,

PR #8587, opened 2017-07-14
add psilopsych pin to the map

Psilopsych:
rancho


therealaslum:
Latitude has an extra "t", also I think you need opening and closing dashes instead of periods. Try:

```
+---
+githubHandle: psilopsych
+Latitude: 45.524285
+Longitude: 123.104630
+---
```

PR #8606, opened 2017-07-15
Create stackermts.json

stackermts:
Add stackermts to the map.


stackermts:
There is a terrible lion up your end and an advantage to an enema at once.

PR #8629, opened 2017-07-15
repaired back to original.

sammyrnycreal:
repaired index.html


sammyrnycreal:
the current index.html file was poorly altered. It needs to be reverted back to the origin.

PR #8639, opened 2017-07-16

jyh2005xx:


soharuna:
Hello,
I experienced same issues. you must have gotten the pin path wrong.  Click on the **Details** text on same line with 'File Path" to watch a simple video on how to resolve this.

PR #8689, opened 2017-07-17
Create dpmartin990.json

dpmartin990:
Pin for dpmartin990

trainingteam @githubteacher


_pins/dpmartin990.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: dpmartin990
```
dpmartin990: Aha, here is where you do this.

dpmartin990:
@dpmartin990 Looks OK to me.

Issue #8783, opened 2017-07-20
Add me, too!


gdickison:
Hello! :wave: Please add me as a collaborator to this repo.

PR #8789, opened 2017-07-20
Adding my pin to the map

gdickison:
4th try - hopefully this will work. It may have been caused by a lower-case "h" in "githubhandle" rather than an upper-case "H" ("githubHandle"). We'll see.

@githubteacher 

closes # #8788 #8783 #8785 #8787 #8786


_pins/gdickison.json
```diff
@@ -0,0 +1,5 @@
+---
```
GregJablonski: 1

_pins/gdickison.json
```diff
@@ -0,0 +1,5 @@
...
+githubHandle: gdickison
```
GregJablonski: 2

PR #8807, opened 2017-07-20
Add handles pin to map

jimvencl2:


FernandaOchoa:
@jimvencl2 This change looks awesome 🎉

PR #8933, opened 2017-07-25
Add Mall of Asia's Globamaze pin to the map

gpborla:
Adding Mall of Asia's steel framed globe structure, Globamaze to the map.

closes #8933


gpborla:
failed pull request

PR #8935, opened 2017-07-25
add gpborla's pin to the map

gpborla:
adding Mall of Asia's steel framed structure Globamaze to the map.

closes #8933


_pins/gpborla.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: gpborla
+latitude: 14.535234
+longtitude: 120.984271
```
hectorsector: @gpborla `longitude` is mispelled. If you correct this, the tests should pass!

gpborla: thanks @hectorsector

gpborla: how do I merge my changes to the master branch?

PR #8940, opened 2017-07-25
Add PGTandreww's pin to the map

PGTandreww:
Adding SF to the map!


sskaysi56:
@PGTandreww  congrats

PR #9003, opened 2017-07-26
Add githubstudent's pin to the map - Oradea

shadowvzs:
Add Oradea to map
close #9003

### DIFF
--- a/_pins/githubstudent1.json
+++ b/_pins/githubstudent1.json
@@ -1,0 +1,4 @@
+---
+githubHandle: githubstudent
+latitude: 47.080416
+longitude: 21.913122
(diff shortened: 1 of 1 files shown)


_pins/githubstudent1.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: githubstudent
+latitude: 47.080416
+longitude: 21.913122
```
WillAbides: It looks like you are almost there, @shadowvzs.  You just need to add one line with `---` at the bottom.

shadowvzs: lol, thank zou, i was wondered what i missed

PR #9065, opened 2017-07-28
add harpenau's pin to the map

harpenau:
adding La Crescent to the map!

### DIFF
--- a/_pins/harpenau-LaCrescent.json
+++ b/_pins/harpenau-LaCrescent.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: harpenau
-latitude:43.838862
-longitude:-91.306611
+latitude: 43.838862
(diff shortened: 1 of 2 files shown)


_pins/harpenau-LaCrescent.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: harpenau
+latitude:43.838862
```
crichID: @harpenau you will need to add a space after each `:` in order for your tests to pass!

_pins/harpenau.json
```diff
@@ -0,0 +1,5 @@
+---
```
crichID: @harpenau this file is perfect, so you may want to consider removing the other file since it contains some small syntax issues.

PR #9133, opened 2017-07-31
NolanKuenzi's Coordinates

NolanKuenzi:
@githubteacher 

I added my coordinates! Let's put rural Oregon on the map! 

closes #9133


NolanKuenzi:
Let's put rural Oregon on the map!

PR #9182, opened 2017-08-01
Add anvythots's pin to map

AnvyThota:
Adding SF to the map!

closes#940


madamebuttons:
hey @AnvyThota ! I think you need to add @githubteacher to your comments. Then just sign in with the Travis CI and you'll be good to go! Happy Gitting!

JohnHenryGaspay:
@AnvyThota instead of githubstudent-Concord it should be AnvyThota-Concord see [https://services.github.com/on-demand/intro-to-github/create-a-branch](url) see details on error on file path.

pins/anvythota.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle:AnvyThota
```
crichID: @AnvyThota you will need to add a space after each `:` in order for the checks to pass!

PR #9201, opened 2017-08-02
Add my Tunis's pin to the map !

Salma-Bouzid:
Adding Tunis to the map!

### DIFF
--- a/Tunis.json
+++ b/Tunis.json
@@ -1,5 +1,0 @@
----
-github handle: Tunis
-Latitude : 36.806495 
-Longitude : 10.181532
----
--- a/Tunis.json
+++ b/Tunis.json
@@ -1,0 +1,5 @@
+---
+github handle: Tunis
+Latitude : 36.806495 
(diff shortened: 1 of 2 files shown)


Tunis.json
```diff
@@ -0,0 +1,5 @@
+---
+github handle: Tunis
```
crichID: @Salma-Bouzid you will need to remove the space in `githubHandle`

Tunis.json
```diff
@@ -0,0 +1,5 @@
...
+Latitude : 36.806495
```
crichID: @Salma-Bouzid please change both `Latitude` and `Longitude` to a lowercase `l` and remove the extra spaces before the `:`. You should only have a single space after the `:`.

brianamarie:
@Salma-Bouzid  I made some changes so the tests will pass. Once everything turns green, I'll go ahead and merge this in since it's been a few days and we want to keep the repository clean. 😄 ✨

PR #9141, opened 2017-07-31
Add VictorArias pin to the map

VictorArias24:
Adding SF to the map!

closes #9141

PR #9221, opened 2017-08-03
Adding VictorArias24's coordinants

VictorArias24:


crichID:
@VictorArias24 this looks great, but it is a duplicate of #9141. I will go ahead and close #9141 and we can keep this one instead.

PR #9294, opened 2017-08-06
Add gg3be0 pin to the map

AbelNTU:
Please add me to the map <3

### DIFF
--- a/_pins/gg3be0.json
+++ b/_pins/gg3be0.json
@@ -1,0 +1,4 @@
+---
+githubHandle: gg3be0
+latitude: 25.0418
+longitude: 121.4966


mbrasseau:
Hello. I'm new too; but I noticed you used your branch name in your code, instead of your user name - that might be why your checks failed. Best of luck!

_pins/gg3be0.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: gg3be0
+latitude: 25.0418
+longitude: 121.4966
```
crichID: @AbelNTU you will need to have three dashes `---` at the end of your file to indicate the end of the yaml. If you need any help, please check out this video: https://vimeo.com/201011288

crichID:
@AbelNTU actually, it looks like this is a duplicate of #9295 so let's close this pull request and focus on the other one!

PR #9370, opened 2017-08-08
Mmaguire cec reading ma

mmaguireCEC:
@githubteacher
how did I do?  not very well.

### DIFF
--- a/_pins/mmaguireCEC.json
+++ b/_pins/mmaguireCEC.json
@@ -1,3 +1,3 @@
-githubhandle: mmaguireCEC
-latitude: 42.489
-longitude: -72.061
++githubhandle: mmaguireCEC
++latitude: 42.489
++longitude: -72.061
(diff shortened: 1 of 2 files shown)


mylocation
```diff
@@ -0,0 +1,3 @@
+githubhandle: mmaguireCEC
```
brianamarie: @mmaguireCEC You will need to delete this file.

_pins/mmaguireCEC.json
```diff
@@ -0,0 +1,3 @@
+githubhandle: mmaguireCEC
```
brianamarie: Make sure to include the three hyphens (like this: `---`) on the first line and the last line, just like in the example.

brianamarie: Also, the spelling and capitalization must be exactly like `githubHandle`

PR #9380, opened 2017-08-08
Add to map!

scoronejo:
Add me!

closes #9380

### DIFF
--- a/_pins/scoronejo.json
+++ b/_pins/scoronejo.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: scoronejo
+latitude: 14.5995
 longitude: 120.9842
-latitute: 14.5995
 ---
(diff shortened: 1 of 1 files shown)


_pins/scoronejo.json
```diff
@@ -0,0 +1,4 @@
+---
+githubhandle: scoronejo
```
brianamarie: Make sure the spelling and capitalization is exactly like `githubHandle`.

_pins/scoronejo.json
```diff
@@ -0,0 +1,4 @@
...
+longitude: 120.9842
+latitute: 14.5995
```
brianamarie: I think that latitude needs to come before longitude, just like in the example. Also, make sure the spelling of `latitude` is correct.

brianamarie: At the end you will need to include three more hyphens, `---`.

PR #9400, opened 2017-08-08
add ysavinaw's pin to the map

ysavinaw:
adding zurich to the map

closes #9400

### DIFF
--- a/_pins/ysavinaw.json
+++ b/_pins/ysavinaw.json
@@ -1,0 +1,4 @@
+---
+githubHandle: ysavinaw
+latitude: 47.405074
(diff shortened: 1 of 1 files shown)


_pins/ysavinaw.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: ysavinaw
+latitude: 47.405074
+longitude: 8.538963
```
brianamarie: Make sure to add three hyphens on the last line, line 5, just like at the top and in the example: `---`

AJLandry1000000000:
looks good

PR #9408, opened 2017-08-09
Add kursatfelsen' pin to the map

kursatfelsen:
Adding SF on the map
closes #940

### DIFF
--- a/_pins/githubstudent01.json
+++ b/_pins/githubstudent01.json
@@ -1,0 +1,4 @@
+...
+githubHandle: kursatfelsen
+latitude: 40.286735
+longitude: 29.649916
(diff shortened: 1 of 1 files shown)


_pins/githubstudent01.json
```diff
@@ -0,0 +1,4 @@
+...
```
brianamarie: Make sure these are three hyphens, not three periods. `---`.

_pins/githubstudent01.json
```diff
@@ -0,0 +1,4 @@
...
+githubHandle: kursatfelsen
+latitude: 40.286735
+longitude: 29.649916
```
brianamarie: Also, you will want to add three hyphens to the very end of the file, just like in the example.

kursatfelsen:
Thank you for the help.

PR #9462, opened 2017-08-12
Add vilnite to the Map.

Vilnite:
Add vilnite to the Map.
Addressed to @githubteacher


Bostoncajun:
I think the issue for the fail is that you added a period after json

PR #9489, opened 2017-08-14
Adding Iceland

Res4MeAdmin:
I've gone and collected the latitute and longitude tto add **Iceland** as my favourite holiday location


asiandub1:
Hi Janelle

What time is in Iceland?

PR #9570, opened 2017-08-20
Add me pin

NinoMaj:


Inaruslynx:
Still open

_pins/ninomaj.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: NinoMaj
+latitude: 45
+longitude: 16
```
hollenberry: Almost there @NinoMaj ! You'll want to add three more dashes to line 5, just like on line 1. Please check out the following video: https://vimeo.com/201011288, or reach out again and let us know if you have questions! :tada:

PR #9578, opened 2017-08-21
Add Prema28-India's pin tothe map

Prema28:


_pins/Prema28-India.jason
```diff
@@ -0,0 +1,4 @@
+---
+githubhandle: Prema28
```
hollenberry: Looks like githubhandle should read `githubHandle`. 👍

_pins/Prema28-India.jason
```diff
@@ -0,0 +1,4 @@
...
+Latitude: 12.971599
```
hollenberry: Almost! This should be all lowercase!

_pins/Prema28-India.jason
```diff
@@ -0,0 +1,4 @@
...
+Longitude: 77.594563
```
hollenberry: All lowercase here as well!

hollenberry: Also, don't forget to add three dashes to the bottom of your code.

elizabethcastillo-asu:
Thank you for this helpful case @Prema28  @hollenberry

PR #9594, opened 2017-08-22
Create deepak.json

Deepakjangra7:
Add deepak's pin to map

PR #9597, opened 2017-08-22
Add Deepakjangra7's pin to map

Deepakjangra7:
Adding SF to map!

closes #9596


beardofedu:
@Deepakjangra7 Your code has similar issues to #9594 , if you make this code look similar (with different latitude and longitude) your pull request should pass.

PR #9615, opened 2017-08-23
add c-andres name to the map

c-andres:
Adding Los angeles to the map!


beardofedu:
@c-andres 👋 you just need to modify 2 things about your file and it should work:

1. The file needs to be in the `_pins/` folder.
2. Change `#githubhandle:` to `githubHandle:`

PR #9644, opened 2017-08-24
Add LKCs pin to the map

lkchandler:


beardofedu:
@lkchandler :wave: teacherbot is currently down for an update, so I manually added you to the course. An email should have been sent to the email you signed up with.

PR #9645, opened 2017-08-24
Create gd1035

gd1035:
Will this add the pin to the map?


_pins/gd1035
```diff
@@ -0,0 +1,2 @@
+latitude:  37.2472
```
gd1035: afds

PR #9752, opened 2017-08-31
Adding Cords

Joshua-Stammen:


JoshuaStammen.json
```diff
@@ -0,0 +1,5 @@
+--
+githubHandle: JoshuaStammen
+latitude: 40.6583
+longitude: -111.9231
+ @githubteacher
```
hectorsector: @joshua-stammen you'll have to replace `@githubteacher` with `---`, and move the file into the `_pins` directory. Here's a video that might help: https://vimeo.com/201011288

PR #9904, opened 2017-09-08
Create 3HotPie

3HotPie:


3HotPie:
Added the longitude and latitude.
Training @githubteacher
Closes #9904

PR #10101, opened 2017-09-19
Add schopelas69's pin to the map

schopelas69:
Adding Salem, NH to the map

### DIFF
--- a/_pins/schopelas69.json
+++ b/_pins/schopelas69.json
@@ -1,5 +1,5 @@
 ---
-gihubHandle: schopelas69
+githubHandle: schopelas69
 latitude: 42.788555
 longitude: -71.200891
 ---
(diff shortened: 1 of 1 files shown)


_pins/schopelas69.json
```diff
@@ -0,0 +1,5 @@
+---
+gihubHandle: schopelas69
+latititude: 42.788555
```
crichID: @schopelas69 it looks like you have an extra `ti` in the word `latitude`. Once you fix it, I believe your tests will pass :wink:

_pins/schopelas69.json
```diff
@@ -0,0 +1,5 @@
+---
+gihubHandle: schopelas69
+latitude: 42.788555
```
AstrildW: now it should work :smile:

schopelas69: Thanks! @crichID @AstrildW

PR #10208, opened 2017-09-22
Join the OU Virtual Reality Association!

brysonreece:
It doesn't matter if you're a programmer, artist, architect, biologist, etc! 

[Join OUVR today to get started learning virtual reality development!](http://ouvirtualreality.org/)


chris-tse:
Nice work man!

PR #10245, opened 2017-09-22
Created Hambonaire

Hambonaire:
@githubteacher 
closes issue #10235

### DIFF
--- a/Hambonaire.json
+++ b/Hambonaire.json
@@ -1,0 +1,5 @@
+---
+_githubHandle: Hambonaire
+latitude: 35.207972
+longitude: -97.445694
+---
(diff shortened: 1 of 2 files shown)


thegreeksdidit:
Hey, I think you're missing the --- at the top and bottom of your file, and added an extra _ in front of git

kjduvall:
You have forgotten the three dashes starting and ending. There is a _ in front of your github username as well.

crichID:
@Hambonaire a couple things are happening here. First, you had two files. The tests would continue to fail as long as either of those files was incorrect. I went ahead and removed one so we can focus on getting the other one correct. 

Here is what you will need to do next:

1. Add `---` at the beginning and end of the file.
2. Remove the `_` in front of githubHandle`

If you need help, please check out this video: https://vimeo.com/201011288

PR #10263, opened 2017-09-22
Add perrycn's pin to the map

### LINKED ISSUE #10262: Add me!
Hello! :wave: Please add me as a collaborator to this repo.

perrycn:
Adding Collinsville to the map!

Closes #10262

### DIFF
--- a/_pins/perrycn.json
+++ b/_pins/perrycn.json
@@ -1,0 +1,4 @@
+...
+gethubHandle:  perrycn
+latitude: 38.670327
+longitude: -89.984548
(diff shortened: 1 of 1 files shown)


_pins/perrycn.json
```diff
@@ -0,0 +1,4 @@
+...
```
crichID: @perrycn this should be `---` instead of `...`. You will need another line of three dashes at the end of the file `---`.

_pins/perrycn.json
```diff
@@ -0,0 +1,4 @@
...
+gethubHandle:  perrycn
```
crichID: @perrycn this should be `git` instead of `get`, so all together it would be `githubHandle`. You will also want to remove the extra space after the `:`.

PR #10473, opened 2017-10-05
Ivanoung hongkong

ivanoung:


ivanoung:
@patcheung Good god I'm failing my testes, I tried changing the contents and making a new file but none of them works. Please help.

brianamarie:
Hi @ivanoung! It looks like you're really close. I think the main key is to add a line at the end with `---`, just like at the top. Let me know if you want more help!

PR #10521, opened 2017-10-07
Create jgramp.json

jgramp:
Use imperatives - e.g. use "add", not "added".


jgramp:
@githubteacher We are closing #10521.

PR #10576, opened 2017-10-11
Adding Berlin coordinates

eminSerin:
@githubteacher Just added the coordinates of Berlin.


ingo69:
@eminSerin 
Goodwork. you can get coordinates at googlemaps by 
putting in you address
right click on the pin on the map
then on the bottom of the modal box will be your exact coordinates! :smile:

PR #10627, opened 2017-10-13
Added my pin onto the map

Mohammedzubairkhan:
I added today ,my pin onto the map. closes #1


wordimont:
@Mohammedzubairkhan This looks great! 🎉

PR #11095, opened 2017-10-26
Create aserafin123.json

aserafin123:
adding latitude/longitude


aserafin123.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: aserafin123
```
crichID: @aserafin123 you are very close, you just need to:

- [ ]  add three dashes `---` on the first and last line of the file
- [ ] put the file in the `_pins` directory

If you need help, check out this video: https://vimeo.com/201011288

PR #11141, opened 2017-10-27
Create unique

uniquear:
Added my coordinates.

### DIFF
--- a/_pins/unique
+++ b/_pins/unique
@@ -1,0 +1,6 @@
+
+---
+githubHandle: uniquear
+latitude: 33.423147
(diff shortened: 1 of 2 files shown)


_pins/unique
```diff
@@ -0,0 +1,6 @@
+
```
danieladamharding: Hey, please remove this line to pass.

crichID: @uniquear you will also need to add the `.json` extension to your file for the tests to pass. If you need any help, check out this video: https://vimeo.com/201011288

PR #11142, opened 2017-10-27
Added new pin with my location

natalie-gomez:


_pins/NatalieGomez
```diff
@@ -0,0 +1,6 @@
+
```
crichID: @natalie-gomez you will need to remove this extra line at the top of your file and also add `.json` as the file extension. If you need any help, please check out this video:https://vimeo.com/201011288

PR #11272, opened 2017-10-29
Add githubstudent's pin to the map.

KhanhDangDuy:
Adding Bach Khoa University to the map.

### DIFF
--- a/pins/githubstudent.json
+++ b/pins/githubstudent.json
@@ -2,3 +2,4 @@
 githubHandle: githubstudent
 latitude: 10.773488
 longitude: 106.659780
+---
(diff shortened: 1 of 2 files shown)


Miari:
Add "---" and the and of the file and commit changes. After it hte check will be successful

KhanhDangDuy:
I modified it. Could you check it again? 
Thanks!!

pins/githubstudent.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: githubstudent
```
crichID: @KhanhDangDuy please replace `githubstudent` with your username!

Miari:
Congratulations!

PR #11450, opened 2017-11-06
Adding coordinates

mahmudzayed:
added cordinates for BD

made changes. @githubteacher


_pins/mahmudzayed.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: mahmudzayed
+latitude: 23.813891
```
mahmudzayed: me

PR #11468, opened 2017-11-07
Add Handlepin to the Map

mountaindogmama:


richmertz:
I failed the same checks. Good luck. We'll figure it out.

handlepin
```diff
@@ -0,0 +1,5 @@
+---
+GitHubhandle-mountaindogmama
```
ash-norsidi: this is a wrong json format. it should be 
`githubHandle: yourusername`
just change the hyphen to a colon (like the rest of the file)

hectorsector: @mountaindogmama this is correct ☝️

PR #11550, opened 2017-11-10
Add sidgarud's pin to the map

sidgarud:


_pins/sidgarud.json
```diff
@@ -0,0 +1,5 @@
+---
+githubhandle:sidgarud
+latitude:18.511146
+longitude:73.796610
```
hectorsector: @sidgarud I believe you can fix the errors you're getting with a space after each `:`!

sidgarud: OK! Thanks!

PR #11552, opened 2017-11-10
Create lightcodes.json.

lightcodes:
Change from Detroit, Mi to Hudson, Mi


_pins/lightcodes.json.
```diff
@@ -0,0 +1,5 @@
+---
+githubhandle-lightcodes
+latitude:41.8538
+longitude:-84.3312
```
hectorsector: @lightcodes Great work so far! You'll need to make sure that your code matches exactly the template in the class, including `---` before and after the code:
```
---
githubHandle: YOUR-USERNAME
latitude: YOUR_LATITUDE
longitude: YOUR_LONGITUDE
---
```

Additionally, there shouldn't be a period at the end of your filename.

PR #11875, opened 2017-11-16
Committing my current location

ruikawahara:
Used NASA's GPS map to list my **CURRENT** latitude and longitude *(Nov. 16, 2017)*

@githubteacher


vladyi74:
Hi!

Seems that your file does not have correct structure - check the line above latitude.
It should look like this:
+githubHandle: <your username>

Good luck,
Vladimir

PR #12197, opened 2017-11-29
Create jazel1969.json

jazel1969:
Issue #12194


_pins/jazel1969.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: jazel1969
+latitude: 52.520007
```
jazel1969: This is great change

jazel1969: Thanks

PR #12245, opened 2017-12-01
add srinivasakumar-a pin to the map

srinivasakumar-a:
Add Bengaluru to the map


srinivasakumar-a:
@teacherbot, thanks for adding and walking me through the course

PR #12272, opened 2017-12-02
Adding myself to the map

TheAndiland:


Vishnu-Thirumalai:
Hey, the code should be saved as a .json file, not as a .andres file as you have it saved.

a-vorontsov:
Try and swap the latitude and longitude positions on the code

VikashKothary:
You now have 4 versions of the file. Delete all the old ones. They are causing the tests to fail.

VikashKothary:
It now passes the test but you'll be adding two versions of your pin. You should remove one.

PR #12275, opened 2017-12-02
Adding pin to map

Vishnu-Thirumalai:


_pin/Vishnu-Thirumalai.json
```diff
@@ -0,0 +1,5 @@
+---
```
TheAndiland: Hey, you are missing the s in _pins
I think thats the problem

RubenG123:
Remove the whitespace on the first line?

_pin/Vishnu-Thirumalai.json
Vishnu-Thirumalai: Oh wow, must have pressed backspace too much. Thanks :)

PR #12363, opened 2017-12-06
Create hagengit.json

hagengit:


hagengit:
I have added my lat and long for adding the pin to the map.  What do you thiink @githubteacher.

This pull request closes #12363

PR #12410, opened 2017-12-07
Adding location coordinants

qinze2002:
Thank you for adding my coordinates :

- latitude: 45.400993
- longitude: -71.882429


qinze2002:
Set the file into the wrong folder

PR #12547, opened 2017-12-11
Create BrianWarris-Redmond.json

BrianWarris:
Pinning yet another Redmond pin to the map


BrianWarris:
this pull request is NO GO.

PR #12905, opened 2017-12-30
Add my position to map.

giswangjian:
I added my position to GitHub Teacher. in #12903 .


_pins/giswangjian.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: giswangjian
```
giswangjian: Hello giswangjian 👍

giswangjian: Thanks to myself 💯

PR #13180, opened 2018-01-08
create joeldread.json pin

joeldread:
adding pin within training


joeldread:
@githubteacher closes #13179

PR #13197, opened 2018-01-09
add paololeg's pin to the map

paololeg:
Adding Arg to the map!


donahue5000:
Howdy! Check the spacing on the longitude line ;)

PR #13239, opened 2018-01-10
Delete my json file

georgelam155:
trying out the workflow here to delete my json file.


hectorsector:
@georgelam155 if you'd like to remove your pin this should do it. 👍 ship it

PR #13244, opened 2018-01-10
Adding My Information to the Map

awalnazif:


_pins/awalnazif.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: awalnazif
```
calenrobinette: You should be sure to add "---" (without quotes) on lines 1 and 5 before and after what you have.

Everything else looks great thought! Keep it up! 👍

PR #13260, opened 2018-01-10
Add Saturn229 to the map

saturn229:
Addin Birmingham to the map

### DIFF
--- a/_pins/Saturn229.json
+++ b/_pins/Saturn229.json
@@ -1,5 +1,5 @@
 ---
 githubHandle: Saturn229
-longitude: 42.546701 
+longitude: 42.546701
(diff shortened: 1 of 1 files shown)


_pins/Saturn229.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: Saturn229
+Longitude: 42.546701 
+Latitude: -83.211319
```
hectorsector: @saturn229 just add `---` to line 5 and you should be all set!

saturn229: Alright will do!

cfbsadmin:
Roll Tide!

PR #13316, opened 2018-01-12

eduardocarneiro:
Adding Eduardo Carneiro location to the GitHub Map


_pins/eduardocarneiro.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: eduardocarneiro
+latidude: 23.689925
```
hectorsector: @eduardocarneiro just check your spelling here, it should be: `latitude`

eduardocarneiro:
thanks!

PR #13364, opened 2018-01-15
Paololeg argentina

waynehiggins:
Im not sure what is happening...  I hope someone can read this.


waynehiggins:
Dont have a clue what I am doing... sigh!

dipstar:
Looks like the CI builds have failed. Wonder if there is a way to manually get them going again.

_pins/paololeg-argentina.json
```diff
@@ -0,0 +1,5 @@
+---
+githubhandle: paololeg
+latitude: -26.844606
+longitude :65.225078
```
hectorsector: @waynehiggins check the spacing here! It should say: `longitude: 65.225078`

ZenMindedBiz: @hectorsector Nice catch! 👍 
@waynehiggins Almost pinned to the map!  🎉

ZenMindedBiz:
I wonder if you can use paololeg as your githubhandle for this .json file when it looks like your githubhandle is @waynehiggins; we'll see.

We're all learning here... very cool to try something new to master.  😄

PR #13476, opened 2018-01-18
adding jtorre58 pin

jtorre58:


adgerrits:
Looks like you forgot to put the '---' at the top and end of the .json file?

PR #13763, opened 2018-01-21
add felipeMello pin to the map

FelipeMello:
add me to the Map!


BhacSsylan:
your filename has to be _pins/username.json

FelipeMello:
Add felipeMello to the map!

close #940

PR #13830, opened 2018-01-23
Create NicMcPhee.json

NicMcPhee:
This adds my latitude and longitude for the University of Minnesota, Morris, for @NicMcPhee. Thanks to @githubteacher for making this possible.

### DIFF
--- a/_pins/NicMcPhee.json
+++ b/_pins/NicMcPhee.json
@@ -1,0 +1,5 @@
+---
+githubHandle: NicMcPhee
+latitude: 45.592690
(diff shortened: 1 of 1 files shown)


_pins/NicMcPhee.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: NicMcPhee
+latitude: 45.592690
+Longitude: -95.899416
```
NicMcPhee: I inadvertently capitalized "Longitude", which causes the build to fail.

PR #14140, opened 2018-01-30
Add katerinaziot's pin on the map

katerinaziot:
**_Adding Davis, CA to the map!_**


toon159:
Yay!  Because of adding more than 1 file.

PR #14314, opened 2018-02-04
Add SwiftDSG's pin to the map

SwiftDSG:
Adding GRS to the map


Fabadiculous:
@SwiftDSG This looks great. 🎉 👍

PR #14458, opened 2018-02-08
Add Bauke777's pin to the map

Bauke777:
Adding Amersfoort to the map!

closes #14458


FJvdPol:
Provide feedback :loudspeaker: and motivation :tada:.

PR #14500, opened 2018-02-09
Create Mozmitch.json

Mozmitch:
Add my location to map using  instructions from @githubteacher . I think this is a cool way to proceed.


Mozmitch:
I like what I see, so far so good

PR #14775, opened 2018-02-15
add pin handle to map

mpalmisano:
adding summit to map


toon159:
change githubhandler to githubHandler :smile:

PR #14777, opened 2018-02-15
Create groove77.json

groove77:
Add groove77's pin to the map


japatti:
@groove77 if you add your latitude and longitude, the status checks will clear

stevensond:
You can find your lat/long coordinates here. https://mynasadata.larc.nasa.gov/latitudelongitude-finder/

Just type them into the comments and it'll get committed.

PR #14811, opened 2018-02-17
Create sm89.json

smandeep89:
@githubteacher what do you thing ?


jwismer:
Looks good. :+1:

_pins/sm89.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: smandeep89 
+latitude: 41.2917
+longitude: -96.1711
```
Amir345: Longitude should not be in negative i think

AnnaDore:
@smandeep89 It looks great!

PR #14831, opened 2018-02-17
Add him's pin to the map

himanshu1603:
Adding LKO to the map


_pins/him.json
```diff
@@ -0,0 +1,5 @@
+---
+him: himanshu1603
```
saladfingerss: needs to say githubHandle: before himanshu1603

PR #14954, opened 2018-02-21
Add student pins to map

Ramwankhede:
abcdefgh


StephenPacker:
@Ramwankhede looks aight 🎉

RoloCao:
You missed the _ in the pins part it has to be _pins instead of just pins @Ramwankhede

AnnaDore:
@Ramwankhede It's a great idea!

AnnaDore:
@Ramwankhede it's a good decision!

PR #15261, opened 2018-03-02
aigulmal   yakutsk

aigulmal:


don-montgomery:
Your .json file is too busy.  Go back to the instructions and see exactly what is needed, then strip out everything else, it should pass travisCI at that point.  Keep after it, you will succeed!

PR #15423, opened 2018-03-07
Add Johannas coordinates

Johannahok:
I want to add my pin to the map.


_pins/Johannahok.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: Johannahok
+latitude: 59.3333
```
4der: @Johannahok yäää!

PR #15424, opened 2018-03-07
Update gunvor coordinates

gunvora:
I want to add my coordinates to the map


_pins/gunvora.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: gunvora
+latitude: 59.3333
+longitude: 18.05
```
mattiasa1974: @gunvora Great!

PR #15425, opened 2018-03-07
Update Paolas coordinates

Paola191:
I want to add my pin to the map

### DIFF
--- a/_pins/Paola191.json
+++ b/_pins/Paola191.json
@@ -1,0 +1,5 @@
+---
+githubHandle: Paola191
+latitude: 59.3333
+longitude: 18.05
+---


_pins/Paola191.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: Paola191
+latitude: 59.3333
+longitude: 18.05
```
annaawgren: Nice work on the longitude!👌🏻⚡️

_pins/Paola191.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: Paola191
+latitude: 59.3333
```
annaawgren: Awesome latitude numbers!💯

_pins/Paola191.json
Paola191: Gracias Anna!

_pins/Paola191.json
Paola191: Tack Anna!

PR #15426, opened 2018-03-07
My coordinates - Freja

Frejabusby:
I want to add my pin to the map!

### DIFF
--- a/_pins/frejabusby.json
+++ b/_pins/frejabusby.json
@@ -1,0 +1,5 @@
+---
+githubHandle: frejabusby
+latitude: 59.3333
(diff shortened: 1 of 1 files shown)


_pins/frejabusby.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: frejabusby
+latitude: 59.3333
+longitude: 18.05
```
KarinFF: @Frejabusby GOOD JOB! 👍

Frejabusby: THANKS @KarinFF !!!! 😄

PR #15427, opened 2018-03-07
Create annaawgren sthlm pin

annaawgren:
I added my coordinates to sthlm so it it will show on the map✨


_pins/ANNAAWGREN.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: ANNAAWGREN
+latitude: 59.3333
+longitude: 18.05
```
mattiasa1974: @annaawgren Looks great!

PR #15428, opened 2018-03-07
Coordinates for me 

AbbeBlubb:
#15415 
Add my coordinates for the pin on the map

### DIFF
--- a/_pins/AbbeBlubb.json
+++ b/_pins/AbbeBlubb.json
@@ -1,0 +1,5 @@
+---
+githubHandle: AbbeBlubb
+latitude: 59.3333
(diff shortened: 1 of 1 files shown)


_pins/AbbeBlubb.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: AbbeBlubb
+latitude: 59.3333
+longitude: 18.05
```
mattiasa1974: @AbbeBlubb Great!

AbbeBlubb: Thanx for feedback!

PR #15430, opened 2018-03-07
Add eliseofferman coordinates

eliseofferman:
I want to add my pin to the map

### DIFF
--- a/_pins/eliseofferman.json
+++ b/_pins/eliseofferman.json
@@ -1,0 +1,5 @@
+---
+githubHandle: eliseofferman
+latitude: 59.3333
+longitude: 18.05
+---


_pins/eliseofferman.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: eliseofferman
+latitude: 59.3333
+longitude: 18.05
```
Johannahok: @eliseofferman good job!

mattiasa1974: @eliseofferman Looks great!

eliseofferman: Tack!!!

eliseofferman: 👍

PR #15432, opened 2018-03-07
Add karinffs coordinates

KarinFF:
I want to add my coordinates to my pin on the map.


Frejabusby:
@stuven1 Can you have a look at this!? I don't know what @KarinFF is trying to do here....

PR #15433, opened 2018-03-07
Adding my location

Davidlicona:
I want to add my pin to the map.

### DIFF
--- a/_pins/davidlicona.json
+++ b/_pins/davidlicona.json
@@ -1,0 +1,5 @@
+---
+githubHandle: Davidlicona
+latitude: 59.3333
+longitude: 18.05
+---


_pins/davidlicona.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: Davidlicona
+latitude: 59.3333
+longitude: 18.05
```
bas: @Davidlicona this looks great! Must be Stockholm as I have the same coordinates. 👍 🎉

_pins/davidlicona.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: Davidlicona
+latitude: 59.3333
```
AbbeBlubb: @Davidlicona Nice coordinates 🥇

Issue #15420, opened 2018-03-07
Please add to map!


kitoreilly:
Hello! Please add me as a collaborator to this repo. 
Figuring out to add Edmond, OK to the map

kitoreilly:
close  #15434

PR #15434, opened 2018-03-07
_pins/kitoreilly.json

kitoreilly:
come fight me in ikea @githubteacher , *just kidding we dont have one*


kitoreilly:
close #15420

PR #15459, opened 2018-03-07
Add don-montgomery's coordinates to map

don-montgomery:
May involve a change from less precise coordinates added earlier via different interface (different interface was a non-code webpage form).

Pull request is made to training team via @githubteacher .

Cannot find an issue number associated with this task.


don-montgomery:
close  #15459

PR #15542, opened 2018-03-10
Add githubestudent's pin to the map

stefanostefanostefano:
Adding stefano to the map


_pins/githubstudent-manchester.json
```diff
@@ -0,0 +1,4 @@
+---
+githubHandle: githubstudent-manchester
+latitude: 53.480759
+longitude: -2.242631
```
AndreasRHoffmann: Looks good. Great job.  👍

PR #15567, opened 2018-03-12
lat and long

htueko:


roseofgold:
it looks like you're missing the dashes at the start and end of the json

_pins/htueko.json
```diff
@@ -0,0 +1,3 @@
+githubHandle: htueko
```
nC3rtainity: Add --- before this line.

PR #15625, opened 2018-03-14
Add Reno to the map!

NahomRasta:
Adding Reno to the map!

closes #940


rikku-pp:
Hello! I think "githubhandle" should be "githubHandle"

PR #15669, opened 2018-03-14
Add michlee-msft's ping to the map

mikelee-msft:
Adding Attleboro to the map!


_pins/mikelee-msft.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: mikelee-msft
```
janechuk: Great job 👍

PR #15747, opened 2018-03-15
Create GAONYANG.json

gaonita:
to add my pin on the map!


gaonita:
I want to **add** my pin to the map! @githubteacher

Laszcody:
Hi, there is no .json extension on one of the files. You should delete it. You also need a 6 digit longitude, like 18.063240 (that is in Stockholm. Keep up the good work :)

PR #15769, opened 2018-03-15
This is my branch

ajantabisht:
@githubteacher what do you think


JuleCvet:
You should write exactly coordinates and add after git/_pins/
 ---
githubHandel: UserName
latitude: 59.3333
longtitude: 18.05
---

PR #15796, opened 2018-03-16
Create _add_pin

MrsCelia:
add pin to map


biplavtimalsina:
Hey there,
You need to add .json extension while you create the file, and also add /_pins in front of the name!
I did the same mistake on my first trial.

Good Work, 
Biplav

PR #15854, opened 2018-03-17
Create juki.xml

JuleCvet:
Exercise again


adilparwez:
you file needs to end with .json instead of .xml

PR #15894, opened 2018-03-19
Create benstott-eduserv.json

benstott-eduserv:
Adding user location
**Markdown Bold**
*Markdown italic*


maju3632:
@benstott-eduserv 
This looks perfect, keep going

PR #15954, opened 2018-03-21
Adding MarieW10 coordinates

MarieW10:
Adding my coordinates.

Making a pull request.

@githubteacher

close #10886


MarieW10:
It would be great to make a similar course like this with the command line thingy

PR #15973, opened 2018-03-21
Create daveb7114.json

daveb7114:
committing my location at Indianapolis, Indiana for this training class


_pins/daveb7114.json
```diff
@@ -0,0 +1,5 @@
+----
+githubhandle: daveb7114
```
sagansdoom: use camel case >> githubHandle

PR #15977, opened 2018-03-21
Adding jlawton-miller coordinates

jlawton-miller:
@jlawton-miller what do you think?

this closes #15972

PR #15979, opened 2018-03-21
Merge pull request #15977 from githubschool/jlawton-miller

jlawton-miller:
Adding jlawton-miller coordinates


jlawton-miller:
adding coordinates

PR #16304, opened 2018-03-30
Add Poojampkrs pin to the map

Poojampkrs:
Adding SF to the map!


jtralongo:
Hi, your file was not in the pins directory and also handle was not capitalized in your script.

CwuAl:
@jtralongo  Thank you for correcting it!!

PR #16356, opened 2018-03-31
add casterivo's pin to the map

casterivo:
adding Kampala to the map!
closes #16356


tdkxc6:
@casterivo good luck on your future endeavors 👍

PR #16363, opened 2018-03-31
Create Jgorsick.json

Jgorsick:
Add my pin to the map!


Jgorsick:
Very hard for me to follow, I had to keep multiple windows open. You might want to suggest that up front. Additional note, I completed this on an IPad it may have been much easier on my laptop. Thanks for making it available for the general public.

PR #16428, opened 2018-04-02
location

pdxgdxcd:
where i be


m1anas:
Rename your file without the last fullstop at the end after "json". Like this:
 _pins/pdxgdxcd.json

smithmich12:
just add .json to the end of your file name

PR #16474, opened 2018-04-03
Adding ginacodes's coordinates

ginacodes:

This pull requests closes #16474.


Rashair:
📢 You should mention @githubteacher 
🎉 And i think there should be only numbers in you latitude and longitude

sroger-git:
Good Job!
this is the template of the file: 
+githubHandle:  your-username-here
+latitude: 37.751
+longitude: -97.822

PR #16488, opened 2018-04-03
added pedreror1 data

pedreror1:
mexico city locations


sfox23:
@pedreror1  try adding a space between the colon and the latitude. Nice work!

PR #16495, opened 2018-04-03
Add laroucol pin to the map

laroucol:


asylbek09:
@laroucol This looks great ! 🎉

PR #16656, opened 2018-04-06
Create location jenshine.json file

jenshine:
@githubteacher, This is my commit. what do you think?


Schiavi13:
@jenshine  You need to delete trailing spaces

PR #16743, opened 2018-04-10

xxolala:
Adding Szczecin to the map!

closes #16743


_pins/gitthubstudent.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: githubstudent
+latitude: 53.457033
+longitude: 14.538708
+'''
```
andrejkurusiov: Unless you're testing, you should'n have added triple-apostrophes at the end. Instead, triple minus shall be there :)

PR #16747, opened 2018-04-10
Add my pin to the map

Piotrek2231:
Addsing me @githubteacher


akbarratt:
There are a couple things you need to do before moving on. First, add @githubteacher in your comment. You also need to use the word "closes" followed by the number of the issue you started in the first step. Add those both to the comment box and you should be good to go. 😄

PR #16859, opened 2018-04-13
Create Katrina's pin

katrinamklawe:
T5H 1T8


BreRobusto:
@katrinamklawe 
It looks like your file isn't in the correct path. All you need to do is go to **Files changed** then click on the pencil icon and change your file path to open-enrollment-classes-introduction-to-github/_pins/Katrina.json
This should help you pass the two checks!

Then change the content of the file to your coordinates.

PR #17095, opened 2018-04-20
Lisle

gkarnik:
**Added gkarnik's pin** @ 41.809882, -88.112388
@githubteacher
closes #3


twk3426:
Awesome pin @gkarnik!!

PR #17149, opened 2018-04-23
Add Vishal5794's pin to map!

vishal5794:
Finally adding Vishal5794!

closes


wikiabhi:
please remove duplicated, you have to add your coordinates only once.

PR #17167, opened 2018-04-23
Add SilasKenneth's pin to the map

SilasKenneth:
Adding Rongo, Kenya to the map


Maurice1989:
Looks good, Well done 😄

PR #17238, opened 2018-04-24
Create birtem coordinates

birtem:
What do you think @mberasategi


mberasategi:
Aha! So it was that tiny little insignificant point there at the end of the file 😉 Congratulations!

PR #17291, opened 2018-04-25
Adding richsharman's coordinates

richsharman:

This is pull requests closes #17291.


richsharman:
add a new pin

sandeepme:
It looks like you need to add .json to the end of the file and you should be good. Good luck!

Schiavi13:
@richsharman You need to add .json at the end of the path filte like this: _pins/richsharman-Longmont.json

PR #17428, opened 2018-04-30
Add cor138 pin to the map

cor138:
Nu is het goed


cor138:
Used subversion 10 years ago. Seems to me Github is more difficult for me.
Good to learn a or maybe the way to use this tool.
Cannot remember the flow but I think I try it once moer before actual use.

Thank you
Cor van Dijk

PR #17441, opened 2018-04-30
Add karoyakani's pin to the map

karoyakani:


karoyakani:
@githubteacher 
 close #17441

PR #17455, opened 2018-05-01
add jenniferhsueh's pin to the map

jenniferhsueh:
Adding vancouver to the map!


jenniferhsueh:
@githubteacher why are 2 of the checks not successful?

samuelferreira00:
@jenniferhsueh Está muito fixe

PR #17555, opened 2018-05-05
Add ygautomo's pin to the map

ygautomo:
Adding location to the map


ygautomo:
Merging ygautomo's pin to the map to the master

PR #17607, opened 2018-05-07
Add AdiPagigih's pin to the map

AdiPagigih:
Adding l to the map!


chuck-it-all:
@AdiPagigih Looks great! 🎉

PR #17675, opened 2018-05-10
Add Ferguson's pin to the map

NFerguson07:
Adding PH to the map!

closes #940


ijjtseng:
it appears you have misspelled githubHandle as `githubHabdle`

0zxo:
@NFerguson07

PR #17704, opened 2018-05-10
Add jlucero520's pin to the map

jlucero520:
Adding Philippines to the map.


jlucero520:
@githubteacher Hi, may I know what I am doing wrong here?

PR #17737, opened 2018-05-11
add githubstudent-newyork's pin to the map

cuny-twong:
add NY to the map


0zxo:
@cuny-twong what do you think?
i added some spaces and now is ready to merge
this pull request closes #17736

PR #17771, opened 2018-05-13
Create raakk1.json

raakk1:
I think the course has to many students to handle.


raakk1:
test but what am I doing

t-swan:
@raakk1 - might be. I got an error message at some point that said a file was truncated (or not saved) because a cap of 1000 had been reached? Not sure if it changed anything, but definitely threw me off for a bit.

0zxo:
@raakk1

PR #17786, opened 2018-05-14
add playalongjake pin to map

playalongjake:
put la on the map


conradwcd:
@playalongjake Looks good! 🎉

PR #17994, opened 2018-05-20
Add the LostAccount handle to the map

LostAccount:
Adding LostAccount to the map.

## IttyBitty


_pins/LostAccount.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: LostAccount
+latitude: 52.133373
+longitude: 21.225365
```
DJBellisle: All coordinates have to be even numbers. Or not. 😕

PR #18024, opened 2018-05-21
Orbroncofan salem

OrBroncoFan:
Adding Salem OR to the map


0zxo:
@OrBroncoFan @githubteacher what do you think?

this pull request closes #18027.

PR #18226, opened 2018-05-29
Adding saarthak-jain's pin to the map

saarthak-jain:
Adding saarthak-jain to pin at Delhi, India


saarthak-jain:
@

saarthak-jain:
Would retry as I was getting 2 unchecked results

PR #18341, opened 2018-06-02
pbuts coords

patrickbutler:
add these coordinates


patrickbutler:
@githubteacher closes #18341

PR #18360, opened 2018-06-04
Add sharanya das's pin to the map

sharanyadas:
Adding cupertino to the map! 

closes #940


0zxo:
@sharanyadas :wave: 
you only need to add "---" at the end,like this:
> \---
githubHandle: sharanyadas
latitude: 37.312362
longitude: -122.029893
--- <<add this
<hr>

and in your previous comment change `#940` to `#18359`

PR #18443, opened 2018-06-05
Adding baktak's coordinates

baktak:
@baktak Thoughts about adding coordinates for our marina?


baktak:
This closes issue #18443

PR #18444, opened 2018-06-05
my coordinates

venkatamujje:

This is to close issue #18444


harshari4299:
@venkatamujje Can you help me out? I am having 2 issues when i want to mark my coordinates on the map.

PR #18439, opened 2018-06-05
Adding harshari4299's Coordinates.

harshari4299:
@harshari4299 What do you think?

So this pull request closes at #18436

PR #18464, opened 2018-06-06
Create harshari4299.json

harshari4299:
This is to Add a pin at my city, New Delhi
@harshari4299
This closes at #18439


0zxo:
@harshari4299 at the end of this page just click close

PR #18476, opened 2018-06-06
Amman

malzarai:
Add Amman to the Map


nparkison:
You just need to change your file extension @malzarai

PR #18494, opened 2018-06-06
Add mochamadrizal26's pin to the map

mochamadrizal26:


drakeshafer:
Looks like you need to move your file to the _pins directory for it to work.

drakeshafer:
I added "_pins" to the beginning of your file name just to test out how editing other people's files worked. I was going to undo it so you could do it yourself, but couldn't figure out how. Oops :sweat_smile:

PR #18563, opened 2018-06-08
Kstirnemann-wellington

kstirnemann:


_pins/kstirnemann.json
```diff
@@ -0,0 +1,5 @@
+---
+githubHandle: kstirnemann
+latitude: 174.776236
+longitude: -41.286460
+---
```
steves09: Excellent job :+1:

PR #18576, opened 2018-06-08
Add HawraMilani's pin to the map

HawraMilani:
Adding London to the map!


diakunwadee:
Did you accept the invitation from teacher bot?
and you should add " close #(your issue). I think

You can try

PR #18644, opened 2018-06-11
add myfj's pin handle to the map

myfj:
Adding SF to the git
#18644


0zxo:
@myfj  :wave:  hallo
i made some changes to your file and now is ready to merge
but the number of your issue is `#18641` instead of `#18644`

:zap:  :cyclone:

PR #18959, opened 2018-06-24
Adding sminar's coordinates

SteveMinar:
@githubteacher How does this look?

1. Added the SteveMinar.json file with boilerplate text
2. replaced the boilerplate with real values


SteveMinar:
@githubteacher - Thanks for the course. It went well and was helpful. One issue: I didn't show up on the map. Any ideas why?

PR #18991, opened 2018-06-25
Create mathurc1.json

mathurc1:
adding coordinates for Hyd


gcastillom90:
I think there is not need to use the degree symbol and the letters N and E in the latitude and longitude.

PR #19082, opened 2018-06-27
Adding coordinates.

emcoffey3:
@githubteacher  What do you think?
This pull request closes #19081.

PR #19086, opened 2018-06-27
Adding Neha's co-ordinates

Neha1214:
@githubteacher What do you think

This pull request closes #19082


Zerosh010:
The first file with the name "NehaP" was failing the test, which is corrected.
Hence, this pull request closes #19086

Zerosh010:
By the way, nice work @Neha1214

PR #19107, opened 2018-06-28
Adding my coordinates

OlaMag:
Second try.


3wcircus:
I'm having the same problem with the walkthrough. Stuck in the same spot and cannot advance to merge. Good luck!

PR #19143, opened 2018-07-01
Add mentisy's pin to the map

mentisy:
Adding Aalesund to the map!


mentisy:
@crichID Do you have any idea why so many pull requests have a failing Travis-CI check? They seem to follow the conventions. Someone suggested checking the travis logs, but I fail to see where they are. I assume they are only to be read by githubschool.

Altaf-Rahman:
@mentisy  Great work. keep moving on

PR #19246, opened 2018-07-05
Add RiverQuinn's pin to the map

RiverQuinn:
Add Joliet to the map!


RiverQuinn:
Is it not possible to get full checks any longer as it stated there was only room for 1000? Can this be updated as people are still using this to learn?

PR #19394, opened 2018-07-12
Adding Crisja04's pin to the map

Crisja04:
Adding pin to the map.

@githubteacher nice tutorial.


surefooted:
@Crisja04 Keep up the good work :100:

PR #19973, opened 2018-08-05
Add miketide08's pin to the map.

miketide08:
Adding Jackson, MS to the map.


miketide08:
Added miketide08's location pin to map.

Thanks @githubteacher  for the help.

closes #19974 
closes #19973

Issue #20093, opened 2018-08-09
Please add me to the list


Deanlee805:
Hello! :wave: Please add me as a collaborator to this repo. 
Hello :)

PR #20094, opened 2018-08-09
Adding Iceland coordinates

Deanlee805:

This closes issue #20093.


NFogell:
Awesome work! :ok_hand:

PR #20247, opened 2018-08-15
Add momedina's pin to the map

momedina:
Adding Kirkland to the map


momedina:
Geez, would it kill you going a little slower :-(

PR #20365, opened 2018-08-22
Add shashanknew47 on map

Shashanknew47:
Adding SF to the map!

closed #940


ChaseMatzinger:
I think _pins is case sensitive - try using a lower case p in your filename :+1:

PR #20552, opened 2018-08-31
Added the pin for my location

bdJohnson72:
Added pin


bdJohnson72:
I added my lat long @githubteacher

PR #20582, opened 2018-09-02
create ergo0-0 coordinates

erg0-0:


tmalamut:
Hi! Make sure your file is in the _pins directory so you can pass the other check.

PR #20940, opened 2018-09-17
Adding GALDRENCYPHR's coordinates

GALDRENCYPHR:
@What do you think

this request closes #940


GALDRENCYPHR:
was pretty awesome and straight forward

PR #21155, opened 2018-09-23
Adding aleloria's coordinants

aleloria:
@githubteacher what about this?


ddh1006:
Good job! You might want to change it from just going into the repository and instead moving it into the _pins folder.

PR #21320, opened 2018-09-27
add DariaUr's pin to the map

DariaUr:
I'm currently adding Milan to the map


mckelveygreg:
uh oh, looks like you didn't add .json to your file, and I'm not sure, but there are some dashes at the end of the section that you may need to copy over?

PR #21658, opened 2018-10-10
Create _pins?TalarekJK17.json

TalarekJK17:
This isn't my home town but it is where i live now


TalarekJK17:
It is just telling about where i live now instead of my actual home town. 
@githubteacher closes #21658

PR #21803, opened 2018-10-13
Create j23evan.json

j23evan:
adding pin location


j23evan:
**Hey** @githubteacher, I have successfully added the Kansas City pin location, and it should be all that is needed for the task. *Also* this closes #21803

PR #21804, opened 2018-10-13
Create linkz715.json/ added my pin

linkz715:
closes##21802


mathieulemieux:
Saw you used DMS coordinates instead of decimal ones! Didn't know I could do that!

Issue #21809, opened 2018-10-13
Hi. Could you kindly add me to this course?


cvalls:
Hello! :wave: Please add me as a collaborator to this repo.

PR #21814, opened 2018-10-13
Raistlin mad Add abd modified 

cvalls:
he metido un comentario al fichero de los gps


cvalls:
hi

reference @githubteacher and @cvalls
this change is to close #21809

PR #22058, opened 2018-10-24
Add heavyslobs' pin to the map

heavyslobs:
Adding Odense to the map


heavyslobs:
I think it was a fun way to get started. There will hopefully be more like this along the learning path - especially about the very basic things like commits, pull requests etc.

